PR #1, opened 2022-04-15
fix: adjustments to generate manifest and filenames correctly on Linux

fell-lucas:
When running `yarn build`, the manifest.json file is generated in the dist folder, but is instantly deleted after the `build` step defined in the `vite.config.ts`, on **Linux**.

It seems the `buildEnd()` function defined in the `plugins/make-manifest.ts` file is being ran at the start of the build instead of at the end.

I made a fix that generates the manifest.json together with the other public files, which seems to then copy it correctly to the dist folder. Since it's now a public file, I thought it would be better to also add indentation to the json.

Chrome also complained about the `dist/Background.js` and `dist/Content.js` filenames starting with capital letters, not letting me import the extension. This PR also fixes that.

And one last thing, when running on watch mode with yarn dev, `fs.watch` can't be set to be recursive on Linux, node doesn't support it. So it's fixed as well.

Feel free to change anything or comment if you have any questions.


nodemon.json
```diff
@@ -0,0 +1,10 @@
+{
+  "watch": [
+    "src"
+  ],
+  "ext": "tsx,css,html,ts",
+  "ignore": [
+    "src/**/*.spec.ts"
+  ],
+  "exec": "vite build"
```
Jonghakseo, line 9: It would be nice to use the delay property to debounce the nodemon settings.

```suggestion
  "exec": "vite build",
  "delay": 300
```

package.json
```diff
@@ -33,6 +33,7 @@
     "eslint-plugin-react": "^7.23.0",
     "eslint-plugin-react-hooks": "^4.3.0",
     "fs-extra": "^10.0.0",
+    "nodemon": "^2.0.15",
```
Jonghakseo, line 36: It seems like a good idea to delete debounces that aren't in use.

Jonghakseo (approved):
Thanks for your contribution~!

Your suggestions are awesome.
In addition to the suggestion, it would be good if two tasks were added.
- removing unused dependencies
- adding nodemon delay

I will do the above two things. 👍🏼

nodemon.json
Jonghakseo: https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/commit/3c86c8864ecfd7afe098d18c384a923fd2d43281

package.json
Jonghakseo: https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/commit/5903e5b6fa6963a2868cceb9428eeeb0108d8939

PR #3, opened 2022-04-15
fix: devtools tab not working

fell-lucas:
Hi, I noticed the chrome dev tools tab wasn't showing up for me, maybe it was also a Linux problem? So I made some adjustments for it to work. Probably not the best fix, but it seems to solve the issue for now.

Also, I have a WIP chrome extension boilerplate using Preact and Windi CSS with Vite, which is heavily inspired by this React one. Is that fine? I'll put credits where its due, thanks for your work!


Jonghakseo:
Thanks for your contribution! 
I didn't know that the panel inside DevTools was not working.
Perhaps my guess is that devtools is not working with esm.

I solved the problem in the commit below. It would be appreciated if you let me know if the issue is resolved after checking.

[fix: DevToolsPanel](https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/commit/aa6423232dffb3e86af33a817a4a9169afb76f3f)

Jonghakseo:
> Also, I have a WIP chrome extension boilerplate using Preact and Windi CSS with Vite, which is heavily inspired by this React one. Is that fine? I'll put credits where its due, thanks for your work!

It is an honour🥇 thank you 👍🏼 
Your project is great. Personally, I am working on a personal extension using this boiler plate.

Thanks for sharing 😄

fell-lucas:
I just tested with your fix commit and it works fine now! I'll close this PR.

PR #15, opened 2022-07-07
fix: run content script with type module

### LINKED ISSUE #13: code in contentView can't call functions on `chrome.runtime` without specifying extension id
Wondering the reason behind you injecting the `contentView` script in the `content` `index.ts` rather than just directly rendering everything in the `content` script? Running into a small issue where I have to manually pass the chrome extension ID in order to run any of the chrome scripts which is just a bit annoying

bowencool:
fix #13

### DIFF
--- a/src/pages/content/index.ts
+++ b/src/pages/content/index.ts
@@ -1,38 +1,6 @@
-try {
-  console.log("content loaded");
-  injectContentViewScript();
-} catch (e) {
-  console.error(e);
-}
-
+console.log("content loaded");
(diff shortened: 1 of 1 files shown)


Jonghakseo (approved):
thanks

rosendolu:
When I have mutilple content scirpt , the output chunck  `modulepreload` became not `inline`, rather using import "xxx";
But chrome extension content scirpt not surpport ,how can i fix it ?

image

image

PR #19, opened 2022-07-08
refactor: content demo

JunyWuuuu91:
write a demo about with content, includes scss、config and js


JunyWuuuu91:
1、`contentView`  is not a crx's component ,so should not create a folder for it
2、make vite chunk build config simple
3、add a demo to show how combining other scss in content

PR #26, opened 2022-08-25
:zap: using internal function

graygalaxy:
using internal function Array.pop() to get the last element of array,
removed getLastElement()

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -43,7 +43,7 @@ export default defineConfig({
           : "assets/js/[name].[hash].js",
         assetFileNames: (assetInfo) => {
           const { dir, name: _name } = path.parse(assetInfo.name);
-          const assetFolder = getLastElement(dir.split("/"));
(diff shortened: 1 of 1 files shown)


Jonghakseo:
@graygalaxy 

Thank you for your contribution.

In my opinion, `getLastElement()` and `Array.prototype.pop()` are functions that do different actions.

I personally prefer the `copy on write` technique as a way to be more predictable and avoid side effects, and `getLastElement()` is the appropriate way to do so. On the other hand, I think the `Array.prototype.pop()` method unnecessarily creates modifications to existing arrays, creating more unpredictable elements in maintenance.

Please give us your opinion or feedback on this part.

Jonghakseo:
No further comments, so I'm closing.

ltyocg:
Using internal function Array.prototype.at() to get the last element of array.
```javascript
[1, 2, 3].at(-1);
// 3
```

PR #101, opened 2023-04-28
Fix #100: ERR_FILE_NOT_FOUND: Race condition during 'npm run dev' when reloading

### LINKED ISSUE #100: ERR_FILE_NOT_FOUND: Race condition during 'npm run dev' when reloading
`utils/reload/initReloadServer.ts` uses `chokidar.watch()` to watch for changes in `dist`.

The problem is that during `npm run build:watch`, the first thing that happens is that all the pre-existing files are unlinked, causing `unlink`, `unlinkDir` and `change` events from `chokidar.watch()`. Then the build process runs and eventually the new files are created causing `add` and `addDir` events.

But if the build takes more the 200ms the unlink events cause `debounce()` to expire causing `ws.send()` to send an update the the client (before the new files have been created). The client calls `window.location.reload()` and we get:

```
Your file couldn’t be accessed
It may have been moved, edited, or deleted.
ERR_FILE_NOT_FOUND
```

image

pmorch:


Jonghakseo:
Thx for your contribution 😃

PR #174, opened 2023-08-20
Support Firefox webextension

### LINKED ISSUE #18: considered web-extension polyfill?
just wondering if you ever considered making this generally compatible with web extension to be used in firefox as well. Seems like most people use https://github.com/mozilla/webextension-polyfill

govza:
closes #18

It looks like Firefox doesn't support adoptedStyleSheets in the content-script, even using not Shadow DOM, but even Document (probably related to this [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1770592)). 

So, regretfully, tailwind should come old way (linking to extension contentStyle) to make it work in FF content script.

IDK is it worth to create workaround to convert and inject twind styles into App.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,10 +9,13 @@
   },
   "scripts": {
     "build": "tsc --noEmit && vite build",
+    "build:firefox": "tsc --noEmit && cross-env __FIREFOX__=true vite build",
     "build:watch": "cross-env __DEV__=true vite build -w",
+    "build:firefox:watch": "cross-env __DEV__=true __FIREFOX__=true vite build -w",
     "build:hmr": "rollup --config utils/reload/rollup.config.ts",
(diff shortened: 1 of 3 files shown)


Jonghakseo (approved):
LGTM

Jonghakseo:
> closes #18
> 
> It looks like Firefox doesn't support adoptedStyleSheets in the content-script, even using not Shadow DOM, but even Document (probably related to this [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1770592)).
> 
> So, regretfully, tailwind should come old way (linking to extension contentStyle) to make it work in FF content script.
> 
> IDK is it worth to create workaround to convert and inject twind styles into App.

Thank you for your work. I'll take a look and leave a comment on the areas that need to be addressed further!

PR #183, opened 2023-09-18
Change color type, to more strict

PatrykKuniczak:
I decide to create it more strict, for better dev experience, IMHO string could be every think in that case :)

And I remove default values, cause it's unnecessary here, because you have 'default' in switch case.


Jonghakseo (approved):
LGTM

Issue #179, opened 2023-09-17
Add typescript support for env file.


PatrykKuniczak:
**Is your feature request related to a problem? Please describe.**
env doesn't have typescript support.
I think good idea is add this:

**Describe the solution you'd like**
Add:

```ts
interface ImportMetaEnv {
    readonly __DEV__: string
    readonly __FIREFOX__: string
}
```

to
```
vite-env.d.ts
```

Jonghakseo:
Currently, the __DEV__ and __FIREFOX__ environment variables are not used in the actual source code.
Therefore, we didn't add them to the environment variable types because we didn't think there was any room to reference them via import.meta.env.

Do you mean in case we use those environment variables inside the source code?

PatrykKuniczak:
You right, i think about it yesterday see the mentioned #187 PR :)

Jonghakseo:
@all-contributors please add @PatrykKuniczak for ideas

PR #187, opened 2023-09-20
Create interface for process env

PatrykKuniczak:
#179 
I think this solution could be helpful for let know typescript, type of env values


Jonghakseo (approved):

Jonghakseo:
@all-contributors please add @PatrykKuniczak for code

PR #196, opened 2023-10-11
feat(error-boundary): create errorBundary and refactor withSuspence

PatrykKuniczak:
I create error boundary and refactor a little wihSupsence, i remove default value on supspence component, cause IMO, if user want to use suspence, he must create a loading component :)


Jonghakseo (approved):
LGTM

Issue #181, opened 2023-09-17
Pre commiter


PatrykKuniczak:
**Is your feature request related to a problem? Please describe.**
Add pre commiter to project.

**Describe the solution you'd like**
Add e.g husky with commitlint, which check commit messages and pre-commiter run eslint and prettier before commit.

Jonghakseo:
If you have a way of thinking, I'd be happy to make it a PR.

PatrykKuniczak:
@Jonghakseo Don't laught i will do that 😆😆😅

PR #197, opened 2023-10-11
Feat/create config for eslint, commitlint, prettier and husky (#181)

PatrykKuniczak:
As i say, i do it #181 

Here's i run this command, and many files have changes, everything works fine

The single quote is my personal preferences, cause less place is wasted, if u think double is better for this boilerplate, you can change it :)

Merge this after #196
I can rebase it and run prettier, then all files have new formatting :)


Jonghakseo (commented):
I don't disagree with anything else, but I personally don't like the useTabs in the prettier setting because I see a lot of unnecessary whitespace when I look at the code on github. I also think it causes a lot of unnecessary filechanges to be logged. For both of those reasons, it would be nice if that item could be modified to false lol.

PatrykKuniczak:
@Jonghakseo You're right, this tabs is unnecessary, i also remove it in my project :)

And i configure wrong, cause eslint have conflict with prettier, now it's good :)

Jonghakseo (approved):
LGTM

Issue #160, opened 2023-07-28
Feature Request: Allow module imports in content scripts


abeledovictor:
I was using a [similar package](https://github.com/lxieyang/chrome-extension-boilerplate-react) for Chrome extension development with react in the past.

That one allows imports in the content script. I think it would be a great feature to add because if not the content script file gets huge, with lots of lines of code and different logic.

Is this something achievable with Vite?

govza:
I'm using the trick actually:
/src/pages/content/index.ts
```js
import("@src/pages/content/Root");
```
and in Root.ts, just doing module imports.

PatrykKuniczak:
@abeledovictor That one is solved or you still looking forward to this update?

abeledovictor:
@PatrykKuniczak I think it is not solved because I cannot use 

`import DefaultImport from 'x'`

or 

`import {NamedImport} from 'y'`

in my content script files.

PatrykKuniczak:
@abeledovictor I will try to reproduce this, and if i have solution i will contact with repo owner for improve this, but repo owner is soo little active.

But i think he want to maintain this repo, cause post some messages last time.

Jonghakseo:
Due to a number of practical issues (career schedule, consulting, personal time), I was not allocating enough physical time to my current project. Fortunately, starting tomorrow, I'll have a long holiday break in Korea, and I'll be spending some of this time with my family before allocating it to repo improvements.

I know I can't realistically address all the issues, but I'll do my best to review them and improve the ones I can :)

If you have any suggestions for issues, please feel free to submit them via PR! I will review them as soon as possible.

abeledovictor:
I do not have the time either. I guess the solution should include adding a rollup plugin like this one @rollup/plugin-commonjs and configuring it to only process the content script file. @PatrykKuniczak

abeledovictor:
> I'm using the trick actually: /src/pages/content/index.ts
> 
> ```js
> import("@src/pages/content/Root");
> ```
> 
> and in Root.ts, just doing module imports.

So, there is only one import that should be done using this syntax? `import("@src/pages/content/Root");` @govza ?

If that is the case, we could change the wording of the comment in `src/pages/content/index.ts`

**From:** 

_Chrome extensions don't support modules in content scripts._


**To:**

_This import is needed only to mount the React Root in the DOM. You can still use module import syntax like this one:

import library from 'x'
import {helperFunction} from 'y'

[...]

PR #246, opened 2023-11-05
Fix: Add use case other modules in content script

Jonghakseo:
- https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/issues/160#issuecomment-1666590141

There is an issue with using import in `content/index.ts` due to module unsupport in the Chrome extension. As a workaround, We can manage dependencies inside other modules that use `import()`. However, this is not visible in the code, making it difficult for users to recognize, so I've added a use case.

Other changes
- Cleaned up dynamic import plugin code


PatrykKuniczak:
@Jonghakseo That's interesting workaround, if i want to run func under condition, i can use this import inside if statement, then this self invoked func is used from e.g 'injected' run when i want it

src/pages/content/injected.ts
PatrykKuniczak, line 1: If theres only 1 func per file, in that case you can use IIFE https://developer.mozilla.org/en-US/docs/Glossary/IIFE

Jonghakseo: Thanks for the nice comments. The reason I didn't use IIFE even though it's a single module is because the file is named `injected.ts`, which I thought was the nature of an integrated module that manages external dependencies.

I thought it was more intuitive to just pick and choose which files I wanted to call inside that file, kind of like `jest.setup.js` :)

Issue #233, opened 2023-10-15
Update packages


PatrykKuniczak:
I think good idea is update all packages, cause some of that is very old e.g rollup, now v4 is available.

I do it on my project, and i have only problem with rollup, rest work's well, i will open PR on days with that, but i don't have now time, to read about v2 -> v4 of rollup, if u have some, u can do it or i will do it in the future, when i close my own project.

Then i want to contribute more ;)

PR #251, opened 2023-11-05
chore/ Update packages

PatrykKuniczak:
#233
I update all packages, but i can't bum vite and @vitejs/plugin-react to latest version, cause on v4 vite don't send to us dir path name of css files, heres the link to issues:

https://github.com/vitejs/vite/issues/12072
https://github.com/rollup/rollup/issues/4724

And without that assetFileNames in vite.config.js don't work properly, and there's not workaround to have that path name, maybe we can try to equal the assetInfo.source to the file source, but it's not optimal, cause the performance on larger file could be bad.

Now we can bundle it into 1 css, but for extenstion it's not work, cause we need separete files for each script.

We can merge this and wait, for this feature on v4, and then we can bump also this, or maybe somebody have workaround for that, then we can use it and bump vite to latest 4.5.0


PatrykKuniczak:
@Jonghakseo It's ready for merge, or no.

You tested it already?

Jonghakseo (approved):
I've tested it under normal development conditions and it seems to be working. 
That took some time.

LGTM

PR #255, opened 2023-11-05
fix/chrome api reference

PatrykKuniczak:
i was remove this namespace, cause it's don't give proper reference to chrome api.

And for react, reference to the chrome api is globally, this nemaspace is not necessary

### DIFF
--- a/src/global.d.ts
+++ b/src/global.d.ts
@@ -1,9 +1,3 @@
-import Chrome from 'chrome';
-
-declare namespace chrome {
-  export default Chrome;
-}
-
 declare module 'virtual:reload-on-update-in-background-script' {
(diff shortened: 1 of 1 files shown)


Jonghakseo:
> i was remove this namespace, cause it's don't give proper reference to chrome api.

I don't understand this sentence. Could you give more details?

PatrykKuniczak:
> > i was remove this namespace, cause it's don't give proper reference to chrome api.
> 
> I don't understand this sentence. Could you give more details?

Let's try to use chrome, before my chane, u don't have reference to any method e.g chrome.window, chrome.runtime.
You have only strange chrome.default but its useless.

But when you remove your custom namespace, then chrome API back to life again :)

Jonghakseo (approved):
LGTM

Issue #257, opened 2023-11-06
Use vitest instead of jest for testing


wonkyDD:
I suggest replacing the testing with **Vitest** instead of jest.    

<br>

**Vitest** guarantees a faster pace of testing and also integrates better with vite, which we are currently using.   
We don't need `jest.confg.js` anymore.

<br>

With vitest, you can remove the `jest`, `jest-environment-jsdom`, `ts-jest`, and `@types/jest` packages
and replace them with `vitest` and `jsdom` for brevity.

<br>

If you don't mind, would you mind assigning it to me?
I've already seen it working fine locally. Thanks for reading, and keep it up!

wonkyDD:
[use-vitest](https://github.com/wonkyDD/chrome-extension-boilerplate-react-vite/tree/use-vitest)

Jonghakseo:
> I suggest replacing the testing with **Vitest** instead of jest.
> 
> 
> **Vitest** guarantees a faster pace of testing and also integrates better with vite, which we are currently using. We don't need `jest.confg.js` anymore.
> 
> 
> With vitest, you can remove the `jest`, `jest-environment-jsdom`, `ts-jest`, and `@types/jest` packages and replace them with `vitest` and `jsdom` for brevity.
> 
> 
> If you don't mind, would you mind assigning it to me? I've already seen it working fine locally. Thanks for reading, and keep it up!

Sure. Can you create a pull request for me?

PatrykKuniczak:
@Jonghakseo If i will finish my own project, i will take a look for that issue, and try to solve that, then i will give info to @wonkyDD  and he can start working again :)

govza:
BTW, using puppeteer/playwright to load the extension into browser and run e2e tests, sometimes comes handy.
Maybe bring them here?

Jonghakseo:
- https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/279

PatrykKuniczak:
@wonkyDD it's merged right now, feel free to start your work again :)

PR #260, opened 2023-11-06
Use vitest

wonkyDD:


.all-contributorsrc
PatrykKuniczak, line 1: I think good idea is disable prettier for this file, cause this bot add new contributor to this file automattically and don't reformat like prettier.

README.md
PatrykKuniczak, line 1: IMO the previouse version is better, cause have short and clear links, paste entire message as a link, isn't good idea.

Create here only the things related to your feature :)

src/pages/devtools/index.html
PatrykKuniczak, line 1: i think it's bug with prettier, but i think why you run prettier for entire project, i was configure lint-staged, for run prettier and eslint only for staged file.

package.json
```diff
@@ -36,14 +36,13 @@
...
     "@types/chrome": "0.0.248",
-    "@types/jest": "29.5.7",
     "@types/node": "20.8.10",
     "@types/react": "18.2.35",
     "@types/react-dom": "18.2.14",
     "@types/ws": "8.5.8",
     "@typescript-eslint/eslint-plugin": "6.9.1",
     "@typescript-eslint/parser": "6.9.1",
-    "@vitejs/plugin-react": "2.2.0",
+    "@vitejs/plugin-react": "4.1.1",
```
PatrykKuniczak, line 45: You can't update this read more in https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/258#issuecomment-1794608629

package.json
```diff
@@ -56,18 +55,17 @@
...
     "npm-run-all": "4.1.5",
     "prettier": "3.0.3",
     "rollup": "4.3.0",
     "sass": "1.69.5",
-    "ts-jest": "29.1.1",
     "ts-loader": "9.5.0",
     "tslib": "2.6.2",
     "typescript": "5.2.2",
-    "vite": "3.2.7",
+    "vite": "4.5.0",
```
PatrykKuniczak, line 67: You can't update this read more in https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/258#issuecomment-1794608629

package.json
```diff
@@ -56,18 +55,17 @@
...
+    "vitest": "^0.34.6",
```
PatrykKuniczak, line 68: I think the good approach is don't use '^' anyway, but here we have deps bot, and they are looking for new versions, as you can see in other packages we don't have it, but here @Jonghakseo could say more if he want this, let's do it :)

package.json
```diff
@@ -56,18 +55,17 @@
     "eslint-plugin-react-hooks": "4.6.0",
     "fs-extra": "11.1.1",
     "husky": "8.0.3",
-    "jest": "29.7.0",
-    "jest-environment-jsdom": "29.7.0",
+    "jsdom": "^22.1.0",
```
PatrykKuniczak, line 58: I think the good approach is don't use '^' anyway, but here we have deps bot, and they are looking for new versions, as you can see in other packages we don't have it, but here @Jonghakseo could say more if he want this, let's do it :)

PatrykKuniczak (changes requested):

PatrykKuniczak:
Don't panic 😁

Let's open new one

wonkyDD:
@PatrykKuniczak 
I should have rebased the upstream commit beforehand and fixed the current branch, but I got screwed.
Thanks for the feedback.

> You can't update this read more in [#258 (comment)](https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/258#issuecomment-1794608629)

To use vitest successfully, we need to upgrade `@vitejs/plugin-react` and `vite`. 
If not, it will encounter this issue https://github.com/vitest-dev/vitest/issues/3283 and tests will continue to fail.

However, after reading the issue you linked to, I realized that I am not able to update it at this time.
I should have read the description of the package update PR beforehand, but I didn't. (apologize for bothering you 🙇)

As you said, I think we should also hold off on [using vitest](https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/issues/257) until the issue is resolved or
there is a workaround.

Thanks a lot.

PatrykKuniczak:
@wonkyDD You no bothered me  😁
If you have small experience with git don't push force, if you no confident your changes, cause you can destroy you work, create also backup branch before rebase, to hqve possibility to recover your work 👻

I was thinking about workaround, i can equal the source of the file, but there's performance on larger files could be sick.

If i have more time on this month, i can try to solve the PR on rollup, then they can relase new version, and then we can pull this and update, as you can see, everything is blocked 😆

Issue #184, opened 2023-09-18
error on code reload


PatrykKuniczak:
**Describe the bug**
Sometimes when i add some more code, e.g 50 lines, then i click inside plugin in chrome, and then i have error(scren shot section):

And i have this maybe once per 10/15 reloads.

Good to know is, i don't close plugin, popup is all time open, and after edit code i click inside plugin.

Almost all times works well, but sometimes don't want to reload.

Maybe it's related to #157 

PS: When project grow up, then i one line change, may lead to an error.

**To Reproduce**
1. Clone and follow all steps for this repo reproduct.
2. Create e.g 1 custm component
3. Try to edit it, or paste several times in `Popup.tsx`

**Screenshots**
![Zrzut ekranu 2023-09-18 225639](https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/assets/64608510/9acfd0b3-a847-4155-a07b-746c598a7601)

For english: `Can't access file`.
File could be moved. Editted or removed.

PS: This info is misleading, because user see, he have some error in plugin, but everything works fine.

I think this console.warn which display that, should be e.g console.info or sth like that, what don't be displayed as error in extenstions section.

![image](https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/assets/64608510/4cdadff9-bcd8-41ed-8e1b-6d28c1dac592)

![image](https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/assets/64608510/a74ddee0-ab38-44d8-a8c2-0e8b13efb6f0)

**Desktop (please complete the following information):**
 - OS: Window 11
 - Browser: chrome latest
 - Node Version: 18.17.1

Jonghakseo:
I'll take a look. :)

PatrykKuniczak:
@Jonghakseo Any new info? :)

Jonghakseo:
@PatrykKuniczak 

Yeah, I can't reproduce it very well. :(

#157 has been resolved, so I think it might have been resolved together, please check.

PatrykKuniczak:
> @PatrykKuniczak
> 
> Yeah, I can't reproduce it very well. :(
> 
> #157 has been resolved, so I think it might have been resolved together, please check.

Ok, after work i take a look :)

PatrykKuniczak:
@Jonghakseo It's still occur

[...]

PR #265, opened 2023-11-11
[Enhance] Modify to let the plugin do the build completion detection not file system.

Jonghakseo:
Fixed https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/issues/184


PatrykKuniczak (changes requested):
Works better, but after rebuild is done, the plugin don't be displaying properly.

THIS HAPPENED ONLY WHEN I'M ADD CODE

Looks like this:
image

I must reopen it manually and then it works, error from #184 don't ocurr, but it's the same case, i must click outside and after that, open again manually and then i can see my changes.

PS: I see also error like this on reload:

THIS HAPPENED ONLY WHEN I'M DELETE CODE

image
image

The entire script is stoped on that error.

Intresting thing is, when i have this small window, from first screen, the code reload is working well, i can remove or add a hundreds line of code, and nothing break, cause plugin don't need to be reloaded, cause it's break already.

If i refresh it manually/reopen, it break again following cases from above

PatrykKuniczak:
@Jonghakseo And the question, we need to create this warn
image

IMO it's a little annoying, if i see i had an error, but in real i don't have, it's only information, clearing it don't work, cause after clean, we use it again, and after each socket disconnect this warn goes into this console.

Jonghakseo:
> IMO it's a little annoying, if i see i had an error, but in real i don't have, it's only information, clearing it don't work, cause after clean, we use it again, and after each socket disconnect this warn goes into this console.

@PatrykKuniczak 
Good. I fixed it to output a log instead of a warning and try to reconnect every 1 second.

Jonghakseo:
> Works better, but after rebuild is done, the plugin don't be displaying properly.
> 
> THIS HAPPENED ONLY WHEN I'M ADD CODE
> ....

In my case, the problem was intermittently clearing files while waiting for a build of an existing file. I've modified the options, can you give it another try?

PatrykKuniczak (approved):
Works perfect <3

PR #308, opened 2023-11-30
feat: add session storage access levels, live update across instances + TSDocs IntelliSense for guidance 

D1no:
Hi thank you for maintaining this boilerplate :) 

## Session Storage Area and Access Levels
I've noticed that `StorageType.Session` wasn't able to be fully utilized, since they need an [extra access-level step](https://developer.chrome.com/docs/extensions/reference/storage/#type-AccessLevel) in chrome to allow content scripts to also access that storage area. A functionality that is pretty important for background workers processing page content.

So I have added that functionality after noticing. 

## In-Code Guidance via TSDocs
Also, since the whole browser extension landscape is rather confusing for most people, I have added brief [TSDocs](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html) to the `createStorage` interface and proximal types to provide in editor guidance:

Kapture 2023-11-30 at 09 39 39

----- 
### _Cross-Page Live Update_
Originally I wanted to extend your creatStorage solution with an aditional `liveUpdate` flag to be able to keep its many instances (content, popup, side panel, etc) synchronized through properly invalidating the cache with [`onChanged` listeners](https://developer.chrome.com/docs/extensions/reference/storage/#event-onChanged)... but I ran out of time for now.

In its current form, the storage and hooks only get pulled once a page starts... kinda omitting the big advantage of using the storage for keeping state in sync. _Like when you change the color of your `theme`, you probably want that to be applied everywhere directly and not have to open and close pages._

Maybe I can do that at another time. All the best from Berlin!

*Note:*

_Heads up: I am not sure how serious the firefox support in a *chrome* extension boilerplate is, but session storage areas are very differently handled by firefox. See here: [chrome.storage.session in firefox](https://stackoverflow.com/questions/76217047/chrome-storage-session-in-firefox)_

### DIFF
--- a/src/shared/storages/base.ts
+++ b/src/shared/storages/base.ts
@@ -72,6 +72,33 @@ type StorageConfig = {
 };
 
 /**
+ * Sets or updates an arbitrary cache with a new value or the result of an update function.
+ */
+async function updateCache<D>(valueOrUpdate: ValueOrUpdate<D>, cache: D | null): Promise<D> {
+  // Type guard to check if our value or update is a function
+  function isFunction<D>(value: ValueOrUpdate<D>): value is (prev: D) => D | Promise<D> {
+    return typeof value === 'function';
+  }
+
+  // Type guard to check in case of a function, if its a Promise
+  function returnsPromise<D>(func: (prev: D) => D | Promise<D>): boolean {
+    // Use ReturnType to infer the return type of the function and check if it's a Promise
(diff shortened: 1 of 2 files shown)


Jonghakseo (approved):
OMG.... it's insane!

I really appreciate the kind help and contributions from the other side of the world.

Jonghakseo:
@all-contributors please add @D1no for code

Issue #307, opened 2023-11-29
How to introduce scss or css in Content UI


webLiang:
<img width="784" alt="image" src="https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/assets/38065991/899c119c-7918-4200-b858-db08cd195fce">

webLiang:
Like this
<img width="804" alt="image" src="https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/assets/38065991/20b07398-89f5-429e-991d-bb96505f1eb9">

webLiang:
Finally discovered this style pour in

<img width="873" alt="image" src="https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/assets/38065991/ed01b782-2289-4f8d-a656-63062388b5c8">

<img width="907" alt="image" src="https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/assets/38065991/54a265d2-2004-4398-a665-cd4ea94ddcb0">

Jonghakseo:
This is a good solution. I'll add it to the code for others to reference!

#313

webLiang:
Thanks   ♪(･ω･)ﾉ

tngflx:
[...]

PR #313, opened 2023-12-02
Enhance: add a css insertion case to shadow dom

Jonghakseo:
- https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/issues/307

I've added the CSS injection method presented in the above issue as an example case.

This should be useful if someone wants to use CSS inside the shadow DOM.


PatrykKuniczak (approved):
LGTM

PR #325, opened 2023-12-08
doc: correct sidepanel chrome version

andronasef:
the written version on the Readme is not correct. 

source: [Chrome Side Panel Documentation] (https://developer.chrome.com/docs/extensions/reference/api/sidePanel#availability)


Jonghakseo (approved):
Oh... you're right! It's my mistake...  thx!

Issue #306, opened 2023-11-27
Can't add a custom Content Script folder


lealahm:
**Describe the bug**
I want to add my own custom content script folder that contains code that only evokes whenever the user clicks on a button in the popup.tsx. I figured out that I had to add the folder path in the **vite.config.ts** file in order to add the folder in the dist directory when build for production. I used the exact same code in the newcontent folder that there is in the content folder ( Changed the import paths ), but whenever I launch the Chrome extension and visit any webpage the default content script won't work and throws up this error:- **Uncaught SyntaxError: Cannot use import statement outside a module**

After much digging, I found out that its mostly because we just can't declare two content script folders in the vite.config.ts file.

**Screenshots**
![error](https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/assets/151762001/384e14fb-d209-47f9-b0f6-2e956576e85f)

**Additional context**
I am not sure what could be the solution to this problem, I just want a default content script from the manifest.js file and another content script folder that injects the scripts based on the user interactions in the popup file or background.ts

PatrykKuniczak:
https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/246/files

That's already fixed, this issue seems to be duplicated with #160

efanibi25:
I couldn't get the previous answer to work so I used this

https://github.com/vitejs/vite/issues/8023
Then modified the package.json

pnpm build:hmr && (run-p wss build:watch build2:watch build3:watch)"

lealahm:
@PatrykKuniczak, Not really, I still couldn't have two different content script folders, one is added on every page through manifest and the other is only added through chrome.executeScript

PatrykKuniczak:
> @PatrykKuniczak, Not really, I still couldn't have two different content script folders, one is added on every page through manifest and the other is only added through chrome.executeScript

It's even possible in pure chrome extension, without React?

lealahm:
@PatrykKuniczak, Yes, its possible with simple vanilla JS project, we can have two different content scrip files, one would be loaded on the every page automatically using manifest and the other only through chrome.scripting.executeScript

PatrykKuniczak:
@lealahm Ok, now i'm understand, you don't want to create second content script, only folder from where you can load next script.

If you have any solution, maybe you want to open PR, and want to work on that, we can help.

I think if you can issue with Vite, that's probably duplicated name issue, let's try to create some suffix.

[...]

PR #329, opened 2023-12-11
inject custom content script from Popup.tsx

lealahm:
In the above code, I try adding a custom content script folder apart from the default `content` folder, which would be the code that can be injected on the page when the user clicks on a button inside popup.tsx or using background script.


src/pages/popup/Popup.tsx
```diff
@@ -9,6 +9,15 @@ import withErrorBoundary from '@src/shared/hoc/withErrorBoundary';
 const Popup = () => {
   const theme = useStorage(exampleThemeStorage);
 
+  const injectContentScript = () {
+    const [tab] = await chrome.tabs.query({ currentWindow: true, active: true });
+    console.log(tab.id)
```
PatrykKuniczak, line 14: Remove this log

vite.config.ts
```diff
@@ -53,6 +53,9 @@ export default defineConfig({
         devtools: resolve(pagesDir, 'devtools', 'index.html'),
         panel: resolve(pagesDir, 'panel', 'index.html'),
         content: resolve(pagesDir, 'content', 'index.ts'),
+
```
PatrykKuniczak, line 56: Remove this blank spaces

PatrykKuniczak (changes requested):
I don't think this is good idea to create duplicated content script.

I mean create step by step instruction, where you describe how to use this, dev which want to implement this, can copy content script, describe only how to attach this to project, like you create in this PR, and describe how to use this excecuteScript :)

src/pages/popup/Popup.tsx
```diff
@@ -9,6 +9,15 @@ import withErrorBoundary from '@src/shared/hoc/withErrorBoundary';
 const Popup = () => {
   const theme = useStorage(exampleThemeStorage);
 
+  const injectContentScript = () {
```
romaindequidt, line 12:
```suggestion
  const injectContentScript = async () => {
```

romaindequidt:
I try to use your branch and had to following erreur as soon as content script is run at the document start : 
image
Have you managed to make it work on your side?

romaindequidt:
May you relate this pull request to its issue #306 , please ? Thanks,

lealahm:
@romaindequidt, Hello, I was also facing the same problem once I tested this out and still trying to solve this problem

src/pages/injectedContent/ui/index.tsx
```diff
@@ -0,0 +1,31 @@
+import { createRoot } from 'react-dom/client';
+import App from '@pages/injectedContent/ui/app';
+import refreshOnUpdate from 'virtual:reload-on-update-in-view';
+import injectedStyle from './injected.css?inline';
+
+refreshOnUpdate('pages/content');
+
+const root = document.createElement('div');
+root.id = 'chrome-extension-boilerplate-react-vite-content-view-root';
```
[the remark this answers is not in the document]
PatrykKuniczak: @lealahm Why did you don't commit this suggestion?

PatrykKuniczak:
@lealahm any moves forward?

tngflx:
@lealahm @romaindequidt the latest vite preload plugin already solved this problem,  https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/commit/685da06fa0d68d86a0d2eeb4fcf8a5fb3f2fb214

aspiers:
I don't think these changes are enough; I was doing something very similar but I also needed to change this line:

https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/blob/685da06fa0d68d86a0d2eeb4fcf8a5fb3f2fb214/utils/plugins/inline-vite-preload-script.ts#L10

In the case of this PR, it would need to be:

```ts
 if (!/content|injectedContent/.test(chunk.fileName)) {
```

src/pages/popup/Popup.tsx
```diff
@@ -9,6 +9,15 @@ import withErrorBoundary from '@src/shared/hoc/withErrorBoundary';
 const Popup = () => {
   const theme = useStorage(exampleThemeStorage);
 
+  const injectContentScript = async () => {
+    const [tab] = await chrome.tabs.query({ currentWindow: true, active: true });
+    console.log(tab.id)
+    chrome.scripting.executeScript({
```
aspiers, line 15: This results in

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'executeScript')
```

I think you need to have the `scripting` permission in the manifest.  You also need to have the `activeTab` permission, or have the content script allowed as described here:

https://developer.chrome.com/docs/extensions/reference/api/scripting

enmanuelmag:
> I try to use your branch and had to following erreur as soon as content script is run at the document start : image Have you managed to make it work on your side?

I facing the same problem, is in this [issue](https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/issues/376), this PR will solve the problem? I want to preserve the behaviour of inject content automatically without open Popup manually. All code works fine when build in devleopment mode, but in production mode I got the mentioned error.

aspiers:
> @lealahm @romaindequidt the latest vite preload plugin already solved this problem, [685da06](https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/commit/685da06fa0d68d86a0d2eeb4fcf8a5fb3f2fb214)

@tngflx It didn't solve it for me; please can you share a working example?

tngflx:
@aspiers Hi man, can I know what's your error? By running the inline script by default in vite.config.ts that should do it, that plugin  is just a hook event inside vite upon bundling.

aspiers:
> Edit : ok I take back my word, the inline script on latest commit is rather buggy, the first line
> 
> ```ts
> if (!/content/.test(chunk.fileName)) {
>               return null;
>      }
> ```
> 
> already f us over...

I already provided this solution in [this comment](https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/329#issuecomment-1913763311).

tngflx:
try my solution 
```ts
async writeBundle(outputOptions, bundle) {
    const outputDir = outputOptions.dir || (outputOptions.file && path.dirname(outputOptions.file));
    let fileToDelete: string | null = null;

    if (!outputDir) {
        console.error('Unable to determine the output directory.');
        return;
    }

    for (const [fileName, outputChunk ] of Object.entries(bundle) as [string, OutputChunk][]) {
        if (!filter(fileName)) {
            continue; // Skip files that don't match the filter
        }

        let modifiedCode = outputChunk.code;

        // Check if the filename contains "index.js"
        if (fileName.includes('index.js')) {
            // Replace import statements with the actual code
            const importStatementRegex = /import\s*\{[^\}]*__vitePreload[^\}]*\}\s*from\s*['"]([^'"]+)['"];/g;
            const exportStatementRegex = /export\s*\{\s*__vitePreload\s*as\s*_\s*\};/g;

            modifiedCode = modifiedCode.replace(importStatementRegex, (_, relativePath) => {
                const absolutePath = path.resolve(outputDir, path.dirname(fileName), relativePath);

                try {
                    let fileContent = fs.readFileSync(absolutePath, 'utf-8');
                    fileContent = fileContent.replace(exportStatementRegex, '');
                    return fileContent;

                } catch (error) {
                    console.error(`Error reading file: ${absolutePath}`, error);
                    return _; // Return the original import statement if there is an error
                }finally {
                    // Store the file path for deletion outside the loop
                    fileToDelete = absolutePath;
                }
            })
        }
        
        try {
            fs.writeFileSync(path.resolve(outputDir, fileName), modifiedCode, 'utf-8');
        } catch (error) {
            console.error(`Error writing file: ${path.resolve(outputDir, fileName)}`, error);
        }

    }

    if (fileToDelete) {
        try {
            fs.unlinkSync(fileToDelete);
        } catch (error) {
            console.error(`Error removing file: ${fileToDelete}`, error);
        }
    }
},
```
I made this solution before i found the latest commit solution, mine works a bit better and skipped the filename check and only focus on index.js

Jonghakseo:
To be honest, I simply started this boilerplate without full knowledge of Vite and the bundling process of Rollup that it uses internally, and I think this is the time to confront that limitation to some extent.

I'm using a lot of vitePlugins to get things to do what I want, or to fix bugs, and I'm increasingly convinced that a fundamental solution is needed.

And that is to completely tear down this boilerplate and fix it at a lower layer based on rollup.

In the process, I think we can also implement a true HMR (not the inconvenient way we have now, where we throw in a refresh method and a path).

The only limitation is my physical time...

aspiers:
@Jonghakseo Makes sense, and I totally understand about the lack of time! So grateful for what you already built here so far though; it enabled me to build my extension which probably would have taken too long for me to do by myself!

PatrykKuniczak:
@Jonghakseo Are you want to implement this PR?

PatrykKuniczak:
@Jonghakseo One more ping :)

Jonghakseo:
> @Jonghakseo One more ping :)

Thanks~! I'll check it soon ^^

PatrykKuniczak:
> > @Jonghakseo One more ping :)
> 
> Thanks~! I'll check it soon ^^

If you decide to move forward with that, maybe i can help with that, it have some months, i need to read it 1 more time :)

PatrykKuniczak:
@Jonghakseo Let's do something with this thread cause maybe somebody waiting for this feature :)

aspiers:
I need this feature :stuck_out_tongue_winking_eye: Would be fantastic if a solution was found :pray:

Jonghakseo:
@PatrykKuniczak @aspiers 

If I understand correctly, this should be worked on top of the legacy branch, is that correct?

Jonghakseo:
Maybe.... this is solution...? 

- #494

PatrykKuniczak:
> Maybe.... this is solution...?
> 
> * [Fix PR #329 #494](https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/494)

I think so

Jonghakseo:
- https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/494

PR #343, opened 2023-12-30
fix: include .tsx, .js, and .jsx files in pnpm lint script

aspiers:
`lint-staged` is already configured to lint not just `*.ts` files but also `*.tsx`, `*.js`, and `*.jsx`.  Also `eslint` is already configured with `eslint-plugin-react`.  So make `pnpm lint` run on the same file types for completeness and consistency.


Jonghakseo (approved):
LGTM

Issue #263, opened 2023-11-09
Add to the Style Library Setup Guide README


Jonghakseo:
**Is your feature request related to a problem? Please describe.**

People have different style libraries they like to use, and adding a specific library to the boilerplate has more downsides than upsides.

**Describe the solution you'd like**
Instead of adding specific libraries to the boilerplate, write a setup guide for each library and add it to the README.
(e.g. Emotion, Chakra UI)

- [x] Chakra UI - https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/270
- [ ] Emotion
- [ ] tailwindcss
- [x] Twind - https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/275

**Additional context**
- https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/issues/243#issuecomment-1793666057

Jonghakseo:
Chakra UI

- https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/270

PatrykKuniczak:
> Chakra UI
> 
> * [docs: add style library guide (Chakra UI) #270](https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/270)

IMO better idea is edit description, and there link PR's, cause it can dissapear when there's a lot of comments.

We don't have many contributors but in larger libs like MUI, all essential info is on Issue/PR description :)

wonkyDD:
@Jonghakseo With your permission, could I add a style guide for [shadcn/ui](https://ui.shadcn.com/)?

Jonghakseo:
> @Jonghakseo With your permission, could I add a style guide for [shadcn/ui](https://ui.shadcn.com/)?

Sure!

2hari:
@wonkyDD can you share some updates ?

PR #346, opened 2023-12-31
add style guide for shadcn ui

wonkyDD:
Related to

https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/issues/263
https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/issues/263#issuecomment-1818199732

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@
 - [Add Style Library](#add-style-library)
     - [Twind](#twind)
     - [Chakra UI](#chakra-ui)
+    - [Shadcn UI](#shadcn-ui)
 - [Pages](#pages)
 - [Screenshots](#screenshots)
     - [NewTab](#newtab)
@@ -355,7 +356,7 @@ createRoot(rootIntoShadow).render(
 
 </details>
 
-### Shadcn UI
+### Shadcn UI <a name="shadcn-ui"></a>
 
 **1. Init tailwind**
 
--- a/README.md
+++ b/README.md
@@ -355,6 +355,73 @@ createRoot(rootIntoShadow).render(
 
 </details>
 
+### Shadcn UI
+
+**1. Init tailwind**
+
+```bash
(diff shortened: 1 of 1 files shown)


Jonghakseo:
I'll test it. Thx!

tngflx:
Hmm twind seems like no longer maintained at all, Im using such script to use tailwindcss straight.. The main problem of using tailwindcss is the difficulty of scoping it into particular element especially during injection content-script. I think my code kinda solve it. Please feel free to improvise.

```ts
const ButtonRenderer = ({ onClickHandler, containerElement, buttonWrapperClasses, buttonName }) => {
    const buttonRef = useRef(null);

    useEffect(() => {
        const shadowRoot = containerElement.attachShadow({ mode: 'open' });

        // Can't directly return buttonElement because we still need to do dom manipulation on this button
        // The only way is to useEffect hook to do dom manipulation before render
        render(
            <div className={buttonWrapperClasses}>
                <button
                    onClick={() => onClickHandler()}
                    className="taoconv_button bg-green-500 hover:bg-green-300 text-black font-bold py-2 px-3 rounded items-center"
                    ref={buttonRef}
                >
                    {buttonName}
                </button>
            </div>,
            shadowRoot
        );


        // Inject the external stylesheet
        const linkElement = document.createElement('link');
        linkElement.rel = 'stylesheet';
        linkElement.href = chrome.runtime.getURL('assets/css/tailwindStyle.chunk.css');
        shadowRoot.appendChild(linkElement);

        // Cleanup logic (optional): Remove the stylesheet and reset buttonRef on component unmount
        return () => {
            shadowRoot.removeChild(linkElement);
        };

    }, [onClickHandler, containerElement]);

    // Return an empty div as a placeholder for the component
    return <div ref={buttonRef}></div>;
};

export default ButtonRenderer;
```

With this code you should be able to run tailwindcss without polluting the globalcss.

PR #396, opened 2024-02-22
Fix type guard to return the correct type

ofir-zeitoun:
if it's Promise, we can use `.then`, `.catch` and `finally`: 
image

otherwise, we can't:
image

Reference: [Using type predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates)


Jonghakseo (approved):
LGTM

Jonghakseo:
@all-contributors please add @ofir-zeitoun for code

PR #407, opened 2024-03-03
fix: check manifest content scripts existence before iteration

barantaran:
Ensure 'forEach' is not called on undefined 'manifest.content_scripts' to prevent the '[make-manifest] Cannot read properties of undefined (reading 'forEach')' error that occurs when the content script section is absent from the manifest.


Jonghakseo (approved):
LGTM

Jonghakseo:
@all-contributors please add @barantaran for code

PR #422, opened 2024-03-21
refactor: add support to non trivial types

ofir-zeitoun:
Add support for non trivial types like Date, Object and Array

for example, Date, (timestampStorage.ts):
```
import { BaseStorage, createStorage, StorageType } from '@src/shared/storages/base';

type TimestampStorage = BaseStorage<Date | undefined> & {
  mark: () => Promise<void>;
};

const storage = createStorage('timestamp', new Date(), {
  storageType: StorageType.Local,
  liveUpdate: true,
  serialization: {
    deserialize: (text: string) => {
      return new Date(text);
    },
    serialize: (value: Date) => {
      return value?.toISOString();
    },
  },
});

export const timestampStorage: TimestampStorage = {
  ...storage,
  mark: async () => {
    await storage.set(() => {
      return new Date();
    });
  },
};
```

Usage:
```
const timestamp = useStorage(timestampStorage);
timestamp?.toISOString()
```

### DIFF
--- a/src/shared/storages/base.ts
+++ b/src/shared/storages/base.ts
@@ -50,7 +50,7 @@ export type BaseStorage<D> = {
   subscribe: (listener: () => void) => () => void;
 };
 
-type StorageConfig = {
+type StorageConfig<D = string> = {
   /**
    * Assign the {@link StorageType} to use.
    * @default Local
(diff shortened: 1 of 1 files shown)


Jonghakseo (approved):
LGTM

Jonghakseo:
@all-contributors please add @ofir-zeitoun for code

PR #430, opened 2024-03-31
Remove npm-run-all dev dependency

hozlucas28:
`npm-run-all` (dev dependency) was replaced by a set of pnpm regexs.


Jonghakseo:
We can remove the npm-run-all dependency, but we want to make it more intuitive for the user to see what we're doing.

PR #444, opened 2024-04-30
fix(blog): variable conflict

webLiang:
image

When I use the main option, the page variable will conflict with the variable of the plug -in and report an error

The improved packing code is as follows:
image


Jonghakseo (approved):
LGTM

Thanks :)

PR #467, opened 2024-05-15
feat/Simplify starting dev environment

PatrykKuniczak:
I was simplified scripts in main package.json

Now it's possible to run 'turbo dev' with run server simultaneously :)


Jonghakseo (approved):
WOW.... LGTM

PR #473, opened 2024-05-18
fix/description of the start scripts

PatrykKuniczak:
I don't change README after #467, now i fix it :)

Also i'm add missing 'alt' for img


Jonghakseo (approved):
LGTM~!

PR #474, opened 2024-05-18
fix/firefox side panel errors on Firefox

PatrykKuniczak:
I was fixed 2 permission errors for Firefox, as i see firefox don't support that side panels :)
image


Jonghakseo (approved):
Oh... I didn't know that... THX :)

PR #490, opened 2024-06-14
 bump turbo version for use new tui feature

jackluson:
Upgrade the turbo version to see the log outputs of each task more clearly.As follows：

image


Jonghakseo (approved):
LGTM

Issue #496, opened 2024-06-16
Build command isn't working after #492


PatrykKuniczak:
**Describe the bug**
After #492 build command is stopping on windows.
Let's rollback this PR or fix it ASAP, cause now entire template is useless on the latest version.

**Screenshots**

![Zrzut ekranu 2024-06-16 164921](https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/assets/64608510/af7476a3-80a9-4843-b533-2cef869a9c15)
![Zrzut ekranu 2024-06-16 164927](https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/assets/64608510/5d8395b3-a178-4d07-982f-a973ccd3ae2c)

**Desktop (please complete the following information):**
 - OS: Window 11
 - Node Version: 20.11.1

Jonghakseo:
Have you tried without `=` into build command? 

```json
"build-dev-server": "pnpm -F hmr build",
```

Jonghakseo:
Well... It looks like permission error.... hmm...

Jonghakseo:
@PatrykKuniczak 

Could you try this? 

- #497

PR #497, opened 2024-06-16
Fix: build error on windows

Jonghakseo:
fix

- #496

It looks like the removal of 'cache:false' to the turbo folder caused a permission error.

- #492

I think the correct fix is to exclude .turbo from the clean target because when someone rerun the turbo build, they might want to use the cache.


PatrykKuniczak (approved):
Now it works ♥

PR #499, opened 2024-06-16
fix: Change folder name from sidepanel -> side-panel

PatrykKuniczak:
As in #498 :)


Jonghakseo (approved):
LGTM

PatrykKuniczak:
@Jonghakseo We're update it in the same time, but you commit it 1 minute before me
image
😂

PR #503, opened 2024-06-16
fix: Add missing build before dev scripts

PatrykKuniczak:
I was restructure a little that scripts and there's a need to place "pnpm build" before 'dev' and 'dev:firefox' scripts, because without if you change e.g package version or sth like that, what causes module rebuild then error occur.
With that build before that's not happened.


PatrykKuniczak:
I need to figure out sth else, cause firefox don't work, let's wait with that PR

PatrykKuniczak:
@Jonghakseo __DEV__ and __FIREFOX__ vars, wasn't loaded into files, seems like cross-env don't work, this PR isn't related to this problem, but here we're need to fix it, cause on firefox plugin can't be loaded.

It works the same way for you?

PatrykKuniczak:
WTF, it isn't work also on the previous versions, i don't know what's going on

PatrykKuniczak:
freaking 
image
we're need to load that env from scripts into files and that should solve entire problem and with that we can simplify this scripts as i done above

Jonghakseo:
@PatrykKuniczak Can you double check this commits?

- https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/503/commits/d6cc142c0ce1b3f07a48002661e7ef7b31e6385b
- https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/503/commits/e885ceb8706e7b4c30c91168b488ce96855befda
- https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/503/commits/67f8ebb307426e53f530fcfca71965342da246ab
- https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/503/commits/f216a74c3acd63793d22f75e3f87e8654de72eab
- https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/503/commits/e060197ca2259ab898aa5aacafa4b549f07b8c45

PatrykKuniczak:
> @PatrykKuniczak Can you double check this commits?
> 
> - https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/503/commits/d6cc142c0ce1b3f07a48002661e7ef7b31e6385b
> - https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/503/commits/e885ceb8706e7b4c30c91168b488ce96855befda
> - https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/503/commits/67f8ebb307426e53f530fcfca71965342da246ab
> - https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/503/commits/f216a74c3acd63793d22f75e3f87e8654de72eab
> - https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/503/commits/e060197ca2259ab898aa5aacafa4b549f07b8c45

Yeah, after dinner 😁

turbo.json
```diff
@@ -1,13 +1,19 @@
 {
   "$schema": "https://turbo.build/schema.json",
   "tasks": {
+    "ready": {
+      "dependsOn": ["^ready"],
```
PatrykKuniczak, line 5: why ready depending on ready?

turbo.json
```diff
@@ -1,13 +1,19 @@
...
+      "outputs": ["dist/**", "build/**"]
+    },
     "dev": {
-      "dependsOn": ["^build"],
+      "dependsOn": ["^build", "ready"],
       "outputs": ["dist/**", "build/**"],
+      "env": ["__FIREFOX__"],
```
PatrykKuniczak, line 11: Where is __DEV__ env, you're completely remove it, why?

Jonghakseo: `cross-env __DEV__=true` is already exist at dev command in each packages.

turbo.json
Jonghakseo: [d83753b](https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/503/commits/d83753bc5950fdf75167d4ff07909673ce746379)

mistake... fixed it!

PatrykKuniczak:
@Jonghakseo For now i can't run e.g 'content' in isolation, cause i can't start that server manually, there's need to be possible to run that server with command like before.
image

Second thing is:
image
Isn't matter i have that package or not, the error occur.
I can't test it because of the case from above(can't start dev server, without run 'pnpm dev' on root).

This __DEV__ env is useless for now in child folders/pages

Jonghakseo:
@PatrykKuniczak 

[`3985a21` (#503)](https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/503/commits/3985a21bda8589d11f4d02bf1637dfb76869417d)

I forgot for a second that `dev-server` should be able to run independently! I've restored it with the above commit.

And the `__DEV__` environment variable is the one used at vite build time, not the one used at runtime, so I think that should be fine?

Jonghakseo (approved):

PatrykKuniczak:
> @PatrykKuniczak 
> 
> [`3985a21` (#503)](https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/503/commits/3985a21bda8589d11f4d02bf1637dfb76869417d)
> 
> I forgot for a second that `dev-server` should be able to run independently! I've restored it with the above commit.
> 
> And the `__DEV__` environment variable is the one used at vite build time, not the one used at runtime, so I think that should be fine? (When we need to use `ENV`, we can use like [this](https://vitejs.dev/guide/env-and-mode))

Yeah i forget, it's vite 😁

PR #513, opened 2024-06-18
build(deps): add lint-staged and husky

wonkyDD:
I tried running `pnpm prettier` and found that formatting has not been applied to the entire repository.
I suggest using husky and lint-staged as  legacy version.


PatrykKuniczak:
And this also can enforce all devs which are working on projects, to have the same structure of code, cause of course sometimes(most of the time XD) we're forgetting to run eslint and prettier before commit

package.json
```diff
@@ -52,6 +55,12 @@
     "typescript": "5.4.5",
     "vite": "^5.2.11"
   },
+  "lint-staged": {
+    "*.{js,jsx,ts,tsx}": [
+      "prettier --write",
+      "eslint --fix"
+    ]
+  },
```
Jonghakseo, line 63: IMO, it's probably a good idea to not give the `eslint --fix` command as a default, as some people don't run eslint very often for performance and other reasons, and it's easy to extend with additional settings.

```suggestion
  "lint-staged": {
    "*.{js,jsx,ts,tsx}": [
      "prettier --write"
    ]
  },
```

Jonghakseo (approved):
LGTM :)

PR #522, opened 2024-06-20
chore: remove logs and update readme

wonkyDD:
Thanks to this https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/503,
it is no longer needed to use `dev-server` and can just use `dev`.

Plus, readme and log have been updated accordingly

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,9 +68,9 @@ This boilerplate is made for creating chrome extensions using React and Typescri
 ### For Chrome: <a name="chrome"></a>
 
 1. Run:
-    - Dev: `pnpm dev` (it's run parallel with `pnpm dev-server` automatically)
-        - When you run with Windows, you should run as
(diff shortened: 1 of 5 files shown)


PatrykKuniczak:
> Thanks to this https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/503,
> it is no longer needed to use `dev-server` and can just use `dev`.
> 
> Plus, readme and log have been updated accordingly

Relating to:
https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/503#issuecomment-2173480146

This script is required regardless, but the part of your PR have sense, removing this console.errors is good.

But in documentation this script still should be, but with new description, relating to running modules in isolation, for default behaviour (run by 'pnpm dev') it's indeed not necessary after #503

Jonghakseo (approved):
LGTM

PR #535, opened 2024-06-26
Fix logo in the popup page

TheShinriel:
# In the Popup page wrong logo is used

If we don't need new-tab page and delete it, the popup page will crash like:
image

Because the logo in the Popup page is used directly from the new-tab page and not its own.

# Solution 

With the fix, the page work alone!
By simply using the existing logo in the Popup folder.
image


Jonghakseo (approved):
LGTM

Issue #526, opened 2024-06-20
`outDir is not inside project root and will not be emptied` warning in vite


wonkyDD:
**Describe the bug**

In each package built with vite, log shows the following warning.
Here is a case for devtools-panel package.

```markdown
(!) outDir chrome-extension-boilerplate-react-vite/devtools-panel/dist is not inside project root
and will not be emptied. Use --emptyOutDir to override.
```

Is it really the right approach to add `--emptyOutDir` to all commands involving Vite?
Given the current repository structure, it seems it might be structurally difficult to eliminate these warnings 🧐

**To Reproduce**

1. git clone
2. pnpm i && pnpm dev

**Expected behavior**

warning disappears during the build.

Jonghakseo:
Fixed~!

- #555

PR #555, opened 2024-07-08
fix: remove messages about empty directory for all modules

PatrykKuniczak:
It's necessary because the build folders are located outside root dir
image

solving #526

### DIFF
--- a/chrome-extension/vite.config.mts
+++ b/chrome-extension/vite.config.mts
@@ -36,6 +36,7 @@ export default defineConfig({
       fileName: 'background',
     },
     outDir,
+    emptyOutDir: true,
     sourcemap: isDev,
     minify: isProduction,
     reportCompressedSize: isProduction,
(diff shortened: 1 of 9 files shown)


chrome-extension/vite.config.mts
```diff
@@ -36,6 +36,7 @@ export default defineConfig({
       fileName: 'background',
     },
     outDir,
+    emptyOutDir: true,
```
Jonghakseo, line 39: The outdir in this config is the `dist` folder, so how about use emptyOutDir to `false` to avoid issues with race conditions?

Actually, I think the problem occurs when I run the build or dev command locally.

```suggestion
    emptyOutDir: false,
```

Jonghakseo (changes requested):
😄

chrome-extension/vite.config.mts
PatrykKuniczak: Yeah, that's need to be false. because of the dist in root, my bad.

Thanks you for the second look :)

Jonghakseo (approved):
LGTM

PR #558, opened 2024-07-08
Feat/project version auto updater

PatrykKuniczak:
I've done it, because nowhere the versions are the same as indeed are in project(0.1.15-current)

But i don't know where i should create a DOCS for this for us, let me know :)

There's inline docs and only you @Jonghakseo udpating project and creating new realeases, if this is good to you, docs isn't necesssary


Jonghakseo:
I think the version of Boilerplate and the product version of the people who actually take Boilerplate and use it are different, so I fixed it at 0.0.1.

And I agree that the documentation needs to be better, I'll see what I can do. :)

PatrykKuniczak:
@Jonghakseo Yeah, but i think that script could be helpful for people which want to bump all versions across project, and for us this could be helpful to stay up to date across all package, after each realase you will do.

Jonghakseo:
> @Jonghakseo Yeah, but i think that script could be helpful for people which want to bump all versions across project
> For us this could be helpful to stay up to date with version across all package.json, after each realase you will do.

You might be right, but I always worry that this boilerplate will become more complicated. Code for convenience, such as this, is one of the reasons I thought so.

Boilerplate is for starting users who have no knowledge about this repository's structure. They might get confused.

So, if this script is not related to the core value of this boilerplate, I think we need more strict and detailed documentation for merge.

PatrykKuniczak:
@Jonghakseo Yeah, it's not related to core, i can do docs for that, gimme a second :)

Leizhenpeng:
I actually have a request, which is to automatically update the manifest version and upload it to Google Market haha, but this requirement is a bit big, and it is worth putting it in a separate pr

PatrykKuniczak:
> I actually have a request, which is to automatically update the manifest version and upload it to Google Market haha, but this requirement is a bit big, and it is worth putting it in a separate pr

Are you want to put this template into Google Market?

PatrykKuniczak:
@Jonghakseo It's ready for re-review :)

Jonghakseo (commented):
https://github.com/user-attachments/assets/8880d492-0d40-450d-aedb-bcc6e8d7d91f


Hmm.... It's not working on my device...

Maybe it depends on OS

PatrykKuniczak:
> 
> https://github.com/user-attachments/assets/8880d492-0d40-450d-aedb-bcc6e8d7d91f
> 
> 
> Hmm.... It's not working on my device...
> 
> Maybe it depends on OS

Yeah, i change shell script type to exe, if i remember right, i need to check what do to, to work on all systems 😊

PatrykKuniczak:
@Jonghakseo Now it should work, i have also added validation for argument

Jonghakseo (approved):
LGTM

I added aliasing in package.json

[feat: add update-version alias in package.json]([58777fe](https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/558/commits/58777fe4221ac4a7c42b5e6d0665c1631cdd96e9))

PatrykKuniczak:
@Jonghakseo Good idea :)

PR #559, opened 2024-07-09
feat: Create base vite config

PatrykKuniczak:


PatrykKuniczak:
@Jonghakseo @Leizhenpeng Import 1 config into another is freaking hard, IDN if i doing sth wrong or what but the build can't be finish.

If you can, pull my changes and try to solve it, i will try to do it on this weekend but it's tricky.

I hope you done it before, and can help :)

PatrykKuniczak:
It's possible, but tricky: 
https://stackoverflow.com/questions/75132236/how-to-share-vite-config-in-monorepo

Jonghakseo:
I think there is a solution using higher order functions... I'll take a look.

PatrykKuniczak:
Ok guys i will do it, as you can see it's necessary to create symlinks to keep .ts extension for files.

I'm exhausted today, i will do it tomorrow, but i now i know it's doable but i need to find the best approach for our purposes :)

PatrykKuniczak:
@Jonghakseo I was place 
```ts
base: ''
```
in base vite, some configs don't have it, this means it was default 
```ts
base: "/"
```
But it seems don't affected this packages, i leave it in that way, if it's neccessary to only some packages have this "" empty, i can refactor it a little, let my know :)

And belive me it's freaking easiest way to implement this, i was trying with .ts and .mts but it's too tricky, it's possible probably but the complexity of this solution is crazzy and we trying to create it more simply, i'm right? :)

PatrykKuniczak:
I see also the version in package isn't current, but leave it for now, after #558 i will change it

PatrykKuniczak:
@Jonghakseo I hope this is enough 😄 

After this big refactor, we need to create docs for entire project 😆

Jonghakseo (changes requested):
- https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/578

I made this PR :)

Plz check above~

Jonghakseo (approved):
LGTM

PR #577, opened 2024-07-15
feat/Create components module

PatrykKuniczak:
I want to create that module, for simplify code, we have redundancy in e.g ToggleButton.

It could be also useful for everybody, cause components will be already defined, and of course everybody using components in React :)

But there's some problems, and i can't configure that
@Jonghakseo Can you run code locally and help me?
Feel free to push changes here, thanks :)


Jonghakseo:
@PatrykKuniczak 

I recently did this on another project. I'll add it as an example :)

Jonghakseo (approved):
@PatrykKuniczak Can you check this local environment?

Jonghakseo:
And also I added new feature about cache busting for package's changing

- https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/585

PatrykKuniczak:
@Jonghakseo You need to fix https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/585#pullrequestreview-2187721895

Because the same error is occuring here 😆

PR #578, opened 2024-07-15
feat: make withPageConfig hoc

Jonghakseo:
I created the vite-config package and cleaned up the code with the hoc pattern. I decided that chrome-extension needed reload rather than refersh, so the benefits of abstraction were small.


PatrykKuniczak (changes requested):
Yeah, it's better implementation ❤️

PatrykKuniczak (approved):
Miss click 🤣

PR #494, opened 2024-06-16
Fix PR #329

Jonghakseo:
### COMMITS

- [feat: add missing permission](https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/494/commits/f488579e337e09dca1788f00c18b39d8c97088d7)
- [fix: check content file test regex](https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/494/commits/a5a07e3055d046545348d1bf57e1c43135924a55)
- [feat: delete injected and add content runtime](https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/494/commits/4008e695311c4b3e059f78e48f246b215b3ea02e)
- [feat: add runtime ui component](https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/494/commits/300095e7cb89e51d1025a081f6e421aec6346ad8)

### VIDEO

https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/assets/53500778/8d86e93f-6b9c-4d81-aa27-dbde4364a4c3

PR #579, opened 2024-07-15
feat/injecting scripts from popup

PatrykKuniczak:
I was done it like in #494 for latest version :)

Don't merge this before #559 , because i want to change sth else after merge this PR.


PatrykKuniczak:
@Jonghakseo IMO after this, we should bump version to 0.2.0, because of new feature

Jonghakseo:
> @Jonghakseo IMO after this, we should bump version to 0.2.0, because of new feature

I agree

Jonghakseo (approved):
LGTM

Issue #571, opened 2024-07-15
 Add ZIP Packaging Command to Scaffold for Chrome Extension Development


Leizhenpeng:
During the development of Chrome extensions, a common requirement is to upload a ZIP file of the project. 

To streamline this process, I propose adding a ZIP packaging command to our scaffolding tools. 

This command should automatically compile and package the extension into a ZIP file, ready for upload. 

This feature would significantly enhance developer productivity by automating a frequently performed task, making the build process more efficient and error-free.

Jonghakseo:
How about github actions?

[build-zip action](https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/actions/workflows/build-zip.yml)

You can download and drag the artifact to upload it to your extension dashboard.

<img width="709" alt="스크린샷 2024-07-15 오후 6 10 06" src="https://github.com/user-attachments/assets/5e3b7799-d433-466a-b311-d396aa032a42">

<img width="423" alt="스크린샷 2024-07-15 오후 6 13 02" src="https://github.com/user-attachments/assets/06eeba00-01ce-4251-8ec7-8cab7472c43b">

Leizhenpeng:
github action is a good sugar But we can't force users to go down this path. 

After all, this needs to be pushed to the remote repository to trigger, and it often takes a long time, and the local package is actually enough for 1-2s

Leizhenpeng:
@Jonghakseo can try this, you only need `pnpm run zip` locally 

By the way, I used to be a high github action enthusiast, but I still like the browser plug-in packaging part, especially when it comes to uploading Google Extension Market

 Remote packaging is not only slow, but also more than once network problems, leaving me speechless

PatrykKuniczak:
@Leizhenpeng But if you want to upload sth to server somewhere, you need to have internet access.

Good idea could be create script which open browser on the [build-zip action](https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/actions/workflows/build-zip.yml)

But you're right in the case where you don't want to push changes, cause you e.g testing sth which don't work on remote but working locally, then your zip script it's amazing feature, after some changes without any commit, you can create quick zip and then upload it anywhere you want 😄

PR #580, opened 2024-07-15
feat: support zip command  #571

Leizhenpeng:


packages/zipper/index.ts
```diff
@@ -0,0 +1,10 @@
+import { resolve } from 'path';
+
+import { zipBundle } from './lib/zip-bundle';
+
+// 打包根目录dist文件
```
PatrykKuniczak, line 5: Create description only in english, for better understanding by everybody :)

packages/zipper/lib/zip-bundle/index.ts
```diff
@@ -0,0 +1,104 @@
...
+      onlyFiles: true,
+    },
+  );
+
+  return new Promise<void>((pResolve, pReject) => {
+    let size = 0;
+    let aborted = false;
+    const timer = Date.now();
+    const zip = new Zip((err, data, final) => {
+      if (aborted) return;
```
PatrykKuniczak, line 71: it's not better to create
```suggestion
      if (!aborted) {rest code ...}
```

Also below you have the same case

packages/zipper/lib/zip-bundle/index.ts
```diff
@@ -0,0 +1,104 @@
...
+
+      if (err) {
+        pReject(err);
+      } else {
+        size += data.length;
+        output.write(data);
+        if (final) {
+          logPackageSize(size, timer);
```
PatrykKuniczak, line 79: Why you don't use data.length directly, this var 'size' seems have no sense, or i don't saw it 😆

packages/zipper/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "@typetrans/tsconfig/utils",
+  "compilerOptions": {
+    "outDir": "dist",
+    "types": ["chrome","node"]
```
PatrykKuniczak, line 5:
```suggestion
    "types": ["chrome", "node"]
```

Jonghakseo (changes requested):
스크린샷 2024-07-18 오후 9 09 31

It makes a zip file with the entire nested folder structure. 🤔

Leizhenpeng:
https://github.com/user-attachments/assets/aea84e72-9f49-4028-bcbd-550e5408a60d

@Jonghakseo I haven't reproduced the bug you said.

Leizhenpeng:
@Jonghakseo It was my negligence. Ha ha, it has been corrected now.

Leizhenpeng:
CleanShot 2024-07-20 at 20 46 33@2x

Jonghakseo (approved):
LGTM

packages/zipper/lib/zip-bundle/index.ts
PatrykKuniczak: @Leizhenpeng Are you eager and able to implement this?

PR #583, opened 2024-07-17
chore/restructure vite config package

PatrykKuniczak:
I think the same structure like other packages could be good, to follow 1 kind of structuring for all packages :)

I have decided also to remove ```moduleProload``` prop, because it's ```true``` for default
Zrzut ekranu 2024-07-17 111050


Jonghakseo (approved):
LGTM

PR #584, opened 2024-07-17
chore/upgrade manifest.js merging approach

PatrykKuniczak:
I want to propose a small upgrade, for simplify merging configs in manifest file :)

### DIFF
--- a/chrome-extension/manifest.js
+++ b/chrome-extension/manifest.js
@@ -1,4 +1,5 @@
 import fs from 'node:fs';
+import deepmerge from 'deepmerge';
 
 const packageJson = JSON.parse(fs.readFileSync('../package.json', 'utf8'));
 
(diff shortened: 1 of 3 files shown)


chrome-extension/manifest.js
```diff
@@ -64,7 +65,7 @@ const manifest = Object.assign(
       },
     ],
   },
-  !isFirefox && { side_panel: { ...sidePanelConfig.side_panel } },
+  !isFirefox && sidePanelConfig,
```
Jonghakseo, line 68:
```suggestion
  sidePanelConfig,
```
We already have logic for FireFox's sidePanel permission, so I think this is unnecessary.

Jonghakseo (commented):
👍🏼

chrome-extension/manifest.js
PatrykKuniczak: OO, in this case, it's completely unnecessary, but i remember there's some warnings in console, without that, i need to check it 1 more time, and i will adjust code 😄

PatrykKuniczak: image

It's still doesn't supported, this check need to be here

PatrykKuniczak: I removed that unnecessary check in config

Jonghakseo (approved):
LGTM

PR #585, opened 2024-07-19
[Enhance] Add cache busting when shared package's changed

Jonghakseo:
One of the frustrations I had with adding UI packages was that modifications to the UI package were not immediately reflected on each page.

Through debugging and looking at the vite source code, I realized that using the build -w(watch) flag overrides the watch setting to true.

This can be circumvented by using a custom watchOption if isDev in withPageConfig, so we implement this.

https://github.com/vitejs/vite/blob/v5.3.3/packages/vite/src/node/build.ts#L719-L734

https://github.com/user-attachments/assets/6cc9d713-3381-4502-a7ef-c85f2a36c03d


PatrykKuniczak (changes requested):
image
image


O mamma mia, hold on with this changes latinos amigos 🤣

Jonghakseo:
@PatrykKuniczak 

Maybe i fiexd it.... 🫠 

[`14557ef` (#585)](https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/585/commits/14557ef7ac53cabd5380f1d14a58c51bf87c2a01)

PatrykKuniczak (approved):
Now it's working 😄

PR #591, opened 2024-07-20
docs: update missing rename

Jonghakseo:


govza:
Can we also rename chrome-extension folder, it's really chromy for firefox developers :)

PR #593, opened 2024-07-20
hotfix/update-version script path

PatrykKuniczak:
I need to change this path, because of error
Zrzut ekranu 2024-07-20 215807

I hope this will works also on Mac OS :)

I changed version to 0.2.2 because i want to realase this as hotfix :)


Jonghakseo:
This isn't work in MacOS and Linux.

https://www.reddit.com/r/node/comments/zy4c67/is_not_recognized_as_an_internal_or_external/

Hmm...

PatrykKuniczak:
@Jonghakseo No it should work, for all OS

But Windows PS and CMD, don't support bash, i have created explanation in DOCS.

If we want to make it working on WIN, we need to add it to PATH.
I was creating a couple scripts like that(With auto adding to PATH), but maybe somebody doesn't want to our template interacting with his/her OS :)

update_version.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/bash
```
Jonghakseo, line 1:
```sh
sh: ./update_version.sh: /usr/bin/bash: bad interpreter: No such file or directory
```

This occur error. Remove this line and it works fine. Is this necessary?

package.json
```diff
@@ -57,7 +59,8 @@
     "tslib": "^2.6.3",
     "turbo": "^2.0.9",
     "typescript": "5.5.3",
-    "vite": "5.3.3"
+    "vite": "5.3.3",
+    "run-script-os": "^1.1.6"
```
Jonghakseo, line 63: 👀

update_version.sh
PatrykKuniczak: Without works also for me, on win and WSL.
I have commited this change

package.json
PatrykKuniczak: This is necessary, because win need to have 'bash' before script name.
IDN why, because if it's outside of package.json script, i can do ./update_version and run this script.
I think it's issue with git bash console, it's unix console, but we don't have unix inside this script runner.
It's probably running in OS related compiler, i think so.

And this is a workaround, to users don't need to use update-version:Them_OS, they only calling update-version with param as default, and this package take responsibility of pick the right one script for them OS.

I hope you understand, what i was trying to explain 😆

PR #594, opened 2024-07-21
fix/node_modules cleaning script

PatrykKuniczak:
I need to add --glob flag.

But after all, error occur because of the lack of node_modules when rimraf is removed by itself xD.

Rimraf suggest to install it globally, i have done postinstall script, which install it automatically with 'pnpm i' :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   },
   "scripts": {
     "clean:node_modules": "pnpx rimraf --glob **/node_modules",
-    "clean": "turbo clean && turbo daemon stop && rimraf dist && rimraf .turbo",
+    "clean": "turbo clean && turbo daemon stop && rimraf dist && rimraf .turbo && pnpm clean:node_modules",
     "clean:install": "pnpm clean:node_modules && pnpm install --frozen-lockfile",
     "build": "turbo build",
     "build:firefox": "cross-env __FIREFOX__=true turbo build",
--- a/package.json
+++ b/package.json
@@ -8,7 +8,9 @@
     "url": "https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite.git"
   },
   "scripts": {
+    "clean:node_modules": "rimraf --glob **/node_modules",
     "clean": "turbo clean && turbo daemon stop && rimraf dist && rimraf .turbo",
+    "clean:install": "pnpm clean:node_modules && pnpm install --frozen-lockfile",
     "build": "turbo build",
     "build:firefox": "cross-env __FIREFOX__=true turbo build",
     "zip": "turbo zip",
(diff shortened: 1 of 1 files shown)


PatrykKuniczak:
@Jonghakseo Maybe add 'clean:node_modules' to 'clean' script ?

PatrykKuniczak:
@Jonghakseo Give me response to the comment above and after this we can merge it 😄

package.json
```diff
@@ -52,7 +53,6 @@
     "lint-staged": "^15.2.7",
     "postcss": "^8.4.38",
     "prettier": "^3.3.3",
-    "rimraf": "^6.0.1",
```
Jonghakseo, line 55: There are tons of scripts that use `rimraf` in other packages as well (`pnpm clean`)

package.json
```diff
@@ -8,8 +8,9 @@
     "url": "https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite.git"
   },
   "scripts": {
+    "postinstall": "pnpm i rimraf -g",
```
Jonghakseo, line 11: Fundamentally, I question whether the ability to clear the node modules folder is a feature that should be added, even at the expense of adding an implicit rimraf dependency.

PatrykKuniczak: @Jonghakseo That's your decision, we can merge this PR or remove this command completely.

package.json
PatrykKuniczak: Yeah, but it's only using in scripts, not in files directly.
image

This global install ```"postinstall": "pnpm i rimraf -g"``` solving this issue.

package.json
Jonghakseo: My obvious solution is to remove the script. It's not a core part of the Chrome extension development, so it's better to remove it than to create noise.

Jonghakseo:
- #620

package.json
PatrykKuniczak: @Jonghakseo It's maybe useful sometimes, but yeah it's not necessary.

package.json
```diff
@@ -8,7 +8,9 @@
     "url": "https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite.git"
   },
   "scripts": {
-    "clean": "turbo clean && turbo daemon stop && rimraf dist && rimraf .turbo",
+    "clean:node_modules": "pnpx rimraf --glob **/node_modules",
+    "clean": "turbo clean && turbo daemon stop && rimraf dist && rimraf .turbo && pnpm clean:node_modules",
```
Jonghakseo, line 12: How about this?

```suggestion
    "clean:node_modules": "pnpx rimraf --glob **/node_modules",
    "clean:turbo": "turbo clean && turbo daemon stop && rimraf dist && rimraf .turbo",
    "clean": "pnpm clean:turbo && pnpm clean:node_modules" ,
```

Jonghakseo (commented):
😄

Jonghakseo (approved):
LGTM

PR #613, opened 2024-07-27
implement i18n package

Jonghakseo:

## Priority*

## Purpose of the PR*

Add I18n Package for internationalize in chrome extension

### Add I18n Package

The i18n package has two main functions.

**First**, it offers the `t`(translate) function, ensuring strong type safety and ease of development for managing multiple languages.

The `t` function behaves exactly the same as `chrome.i18n.getMessage` in production. However, it is implemented as a dev-only function to allow developers to see quick results without reloading the extension.

스크린샷 2024-07-28 오전 1 18 10

This ensures that features like HMR work as expected while also providing strong type safety by inferring the key values of the messages entered in message.json.

**Secondly**, developers do not need to manually update the utility functions for type inference whenever adding or removing languages. Through the `generate-i18n.mjs` file, simply creating or deleting folders and new JSON files is sufficient to prepare support for new languages.

### Add Usage into `NewTab.tsx`

```tsx
const NewTab = () => {
  //...
  console.log(t('hello', 'World')); // Hello, World ('World' is a substitution)

  return (
    <div>
      <header>
        {/*...*/}
        <Button onClick={exampleThemeStorage.toggle} theme={theme}>
          {t('toggleTheme')} // Toggle Theme
        </Button>
      </header>
    </div>
  );
};
```

## How to check the feature

https://github.com/user-attachments/assets/9542dcc8-3b01-4e16-92b0-fc15b8edf9e6

## Reference

https://developer.chrome.com/docs/extensions/reference/api/i18n
https://developer.chrome.com/docs/extensions/how-to/ui/localization-message-formats#placeholders


packages/i18n/build.mjs
```diff
@@ -0,0 +1,31 @@
...
+    target: "es6",
+    outdir: "./dist",
+    sourcemap: true,
+    format: "esm"
+  });
+
+  const outDir = path.resolve("..", "..", "dist");
+  const localePath = path.resolve(outDir, "_locales");
+  rimraf.sync(localePath);
+  fs.cpSync(path.resolve("locales"), localePath, { recursive: true });
```
Jonghakseo, line 26: The locales folder is copied to the output path with only its name changed.

packages/i18n/index.ts
```diff
@@ -0,0 +1,6 @@
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+import { t as t_dev_or_prod } from './lib/i18n';
+import { t as t_dev } from './lib/i18n-dev';
+
+export const t = t_dev_or_prod as unknown as typeof t_dev;
```
Jonghakseo, line 6: To prevent file changes with the `pnpm build` and `pnpm dev` commands, I have set them to be ignored. Type errors are ignored with comments, while type support remains the same as `i18n-dev.ts`.

packages/i18n/genenrate-i18n.mjs
```diff
@@ -0,0 +1,123 @@
...
+  const defaultLocale = Intl.DateTimeFormat().resolvedOptions().locale.replace('-', '_');
+  if (locales.includes(defaultLocale)) {
+    return defaultLocale;
+  }
+  const defaultLocaleWithoutRegion = defaultLocale.split('_')[0];
+  if (locales.includes(defaultLocaleWithoutRegion)) {
+    return defaultLocaleWithoutRegion;
+  }
+  return firstLocale;
+})()`
```
Jonghakseo, line 103: This is a hacky code to provide the default language for the runtime environment. 🫠

packages/i18n/README.md
```diff
@@ -0,0 +1,198 @@
...
+}
+```
+
+If you want to use placeholders, you can use the following format.
+
+https://developer.chrome.com/docs/extensions/how-to/ui/localization-message-formats#placeholders
+
+> [!Caution]
+> You should use placeholders in this format: `$1`, `$2`, `$3`, ... $/d+
+> Still, We don't support the placeholders field in the `messages.json` file.
```
PatrykKuniczak, line 55: Maybe, implement it also, on this PR?
I seeing you have 'TODO' in 'i18n-dev' file

packages/i18n/README.md
```diff
@@ -0,0 +1,198 @@
...
+
+`locales/en/messages.json`
+
+```json
+{
+  "helloAnd": {
+    "message": "Hello, $1 $2"
```
PatrykKuniczak, line 62: I don't understand how to use this placeholders in json i have this $1, as you have described here, but nothing was showed.

OOO now i got it.
```ts
console.log(t('hello', ['hello', 'xd']));
```

I should define array after.
I think good idea could be, to add example like this in this README, because of this developer.chrome example isn't good enough in this case.

packages/i18n/lib/i18n-dev.ts
```diff
@@ -0,0 +1,26 @@
...
+  }
+  return message.replace(/\$(\d+)/, substitutions);
+}
+
+function removePlaceholder(message: string) {
+  return message.replace(/\$\d+/g, '');
+}
+
+export const t = (...args: Parameters<typeof translate>) => {
+  return removePlaceholder(translate(...args));
```
PatrykKuniczak, line 23: I looking for that, and my thoughts about it have destroyed xD

I think substitusion is placeholders.

PLS create more examples and explanation in README 🔥

pages/new-tab/src/NewTab.tsx
```diff
@@ -3,12 +3,15 @@ import '@src/NewTab.scss';
...
 import { exampleThemeStorage } from '@extension/storage';
 import { Button } from '@extension/ui';
+import { t } from '@extension/i18n';
 
 const NewTab = () => {
   const theme = useStorageSuspense(exampleThemeStorage);
   const isLight = theme === 'light';
   const logo = isLight ? 'new-tab/logo_horizontal.svg' : 'new-tab/logo_horizontal_dark.svg';
 
+  console.log(t('hello', 'World'));
```
PatrykKuniczak, line 13: OO there's some example, but as i type above, this should be also in README 😸

PatrykKuniczak (changes requested):
1. pnpm script for generating, should be located in the root(I think good idea is create a pointing script, which run this nested(from i18n) script in root), like other scripts, because until now we was using only root package.json and magic happened in the subfolders :)
I think this should work on the same way.
2. There's no function to change language in real time, i mean by user(switching between languages).
3. Move all 'build.' and 'generate ...' file into 'builders' folder(maybe it should be inside lib?)

After all it's amazing, because it works in all pages, if you want to implement it without turborepo/modules pattern, it's tricky to work properly in background because i was implementing it on legacy version on my project ;)

Jonghakseo:
> 1. pnpm script for generating, should be located in the root(I think good idea is create a pointing script, which run this nested(from i18n) script in root), like other scripts, because until now we was using only root package.json and magic happened in the subfolders :)
>    I think this should work on the same way.
> 2. There's no function to change language in real time, i mean by user(switching between languages).
> 3. Move all 'build.' and 'generate ...' file into 'builders' folder(maybe it should be inside lib?)


1. I agree :)
2. The ability to change the language in real-time is not something that the Chrome extension's multilingual capabilities provide. There are already many great multilingual libraries that provide the ability to change the language at runtime. What we're focusing on here is a bit of support for how multilingualism is handled in Chrome extensions.
3. OK :)


> After all it's amazing, because it works in all pages, if you want to implement it without turborepo/modules pattern, it's tricky to work properly in background because i was implementing it on legacy version on my project ;)

I didn't test the behavior in the background! I'll check the behavior, and it's true that it does rely on the behavior of Turborepo (more precisely on the watch function). 

I thought it would be a good way to do it with less effort :)

pages/new-tab/src/NewTab.tsx
Jonghakseo: I thought I documented it well enough in the README.md in the i18n package, but what do you think?

packages/i18n/README.md
Jonghakseo: I'll look into the way placeholders work a bit more and refine it so we can implement it all at once~!

packages/i18n/genenrate-i18n.mjs
```diff
@@ -0,0 +1,123 @@
...
+
+  const getMessageFromLocaleFile = `/**
+ * This file is generated by generate-i18n.mjs
+ * Do not edit this file directly
+ */
+${locales.map((locale) => `import ${locale}Message from '../locales/${locale}/messages.json';`).join("\n")}
+
+export function getMessageFromLocale(locale: string) {
+  switch (locale) {
+${locales.map((locale) => `    case '${locale}':
```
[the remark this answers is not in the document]
Jonghakseo: When we use dynamic import, we have the constraint that the function must be called asynchronously.

However, the problem you suggest is reasonable and I'll have to think about it. 

I'm currently maintaining a product in my company that supports 15 different languages, and each language file is about 500kB in size, so obviously the issue of optimizing for multiple languages and dynamic importing becomes very important at this scale.

But what if it's a Chrome extension, not a webpage, and the issue only occurs during the development phase...?
I guess it doesn't really matter. 🤔

pages/new-tab/src/NewTab.tsx
PatrykKuniczak: I think, if you implement 'placeholder' filed in json, there's no need to create more clarification, but no it's a little tricky to use it 😆

PatrykKuniczak:
@Jonghakseo It's working good in background in this version :)

2. Yeah, that's true, but i think it could be good feature, because users don't need to download other package only for switching language.

I was implementing language switching on my project some times ago.
https://github.com/PatrykKuniczak/YT_Notifier/blob/main/react-app/utils/hooks/use-language-switch.ts
https://github.com/PatrykKuniczak/YT_Notifier/blob/main/react-app/src/pages/popup/components/language-selector.tsx

I think if it will be build in, this could be good to our users, let's think about it a little 😸

packages/i18n/genenrate-i18n.mjs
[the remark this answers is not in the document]
PatrykKuniczak: It depends of the extension complexity, as you say.
This solution is good, but if you have better idea, try to implement it, sth more generic.

Jonghakseo:
> > 1. pnpm script for generating, should be located in the root(I think good idea is create a pointing script, which run this nested(from i18n) script in root), like other scripts, because until now we was using only root package.json and magic happened in the subfolders :)
> >    I think this should work on the same way.
> > 2. There's no function to change language in real time, i mean by user(switching between languages).
> > 3. Move all 'build.' and 'generate ...' file into 'builders' folder(maybe it should be inside lib?)
> 
> 1. I agree :)
> 2. The ability to change the language in real-time is not something that the Chrome extension's multilingual capabilities provide. There are already many great multilingual libraries that provide the ability to change the language at runtime. What we're focusing on here is a bit of support for how multilingualism is handled in Chrome extensions.
> 3. OK :)
> 
> > After all it's amazing, because it works in all pages, if you want to implement it without turborepo/modules pattern, it's tricky to work properly in background because i was implementing it on legacy version on my project ;)
> 
> I didn't test the behavior in the background! I'll check the behavior, and it's true that it does rely on the behavior of Turborepo (more precisely on the watch function).
> 
> I thought it would be a good way to do it with less effort :)

I've been thinking about it again, 

1. In my user experience, the only thing that will run commands related to the i18n package in the top-level path is `genenrate-i8n', which is already included in the build and ready commands, so I don't think there is any need to run it separately in the top-level path! 😄 

2. There are a lot of things to think about for implementation... First of all, the multilingual capabilities of the Chrome extension are determined by the `browser's language setting`. If I change the language I want to use at runtime, I have to change to an implementation that doesn't rely on `chrome.i18n` API. Does that defeat the purpose of providing the convenience of using the `chrome.i18n` API? I'm not sure. 🫠 

3. I've tried moving the `build.~` and `generate.~` files to builders or into libs, but it feels like unnecessary encapsulation that's hard to understand.

packages/i18n/README.md
Jonghakseo: I've been looking for a library that implements (or at least mimics) Chrome's i18n feature and I can't find one. I've looked at Chromium's source code, but there's only test code for the i18n feature, so it's hard to reference a detailed implementation...

https://chromium.googlesource.com/chromium/src/+/901414aca5338840519c9fa7ee3fb82cb4f0f18f/chrome/common/extensions/api/extension_api.json
https://chromium.googlesource.com/chromium/src/+/901414aca5338840519c9fa7ee3fb82cb4f0f18f/chrome/common/extensions
https://github.com/chromium/chromium/blob/main/chrome/browser/extensions/api/i18n/i18n_apitest.cc

Jonghakseo: https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/613/commits/8abc4f6f07aba0b2912a68f5339b9d33e98b7350

I think this is a implementation for chrom.i18n API.

PatrykKuniczak:
@Jonghakseo 
1. Yeah, right
2. Yeaaah, that's true, i was thinking about users which want to change language for e.g 'english', but then they can change language for entire browser, and respectly plugin also have english. That's a point 😄 
3. Ok, if it's complicating, that's ok to leave it in that shape.

packages/i18n/lib/i18n-dev.ts
```diff
@@ -0,0 +1,44 @@
+import { DevLocale, MessageKey } from './type';
+import { defaultLocale, getMessageFromLocale } from './getMessageFromLocale';
+
+type I18nValue = {
+  message: string;
+  placeholders?: Record<string, { content?: string; exampe?: string }>;
```
PatrykKuniczak, line 6: Not that way?

```suggestion
  placeholders?: Record<string, { content?: string; example?: string }>;
```

PatrykKuniczak:
@Jonghakseo Now users can define 'placeholder' prop in json?
Because i don't see any new examples.

Jonghakseo:
> @Jonghakseo이제 사용자는 'placeholder' prop을 json으로 정의할 수 있나요? 새로운 예가 보이지 않거든요.

I added example on README.md :)

[252277e](https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/613/commits/252277e8f5bffe03bf40f38e5f0b9736b84884a3)

packages/i18n/README.md
```diff
@@ -129,41 +101,91 @@ Just import the `t` function and use it to translate the key.
...
 ### Placeholders
 
-If you want to use placeholders, you can pass the second argument as a string or an array.
+If you want to use placeholders, you can use the following format.
+
+> For more information, see the [Message Placeholders](https://developer.chrome.com/docs/extensions/how-to/ui/localization-message-formats#placeholders) section.
+
+`locales/en/messages.json`
+
+
```
PatrykKuniczak, line 127: 1 additional blank space, relative to example below 😄

packages/i18n/README.md
```diff
@@ -129,41 +101,91 @@ Just import the `t` function and use it to translate the key.
...
+```json
+{
+  "greeting": {
```
PatrykKuniczak, line 130: IMO, it's more important to create example for this key, than for ```hello``` and how to use this 'placeholders'.

Maybe create 2 examples.

packages/i18n/README.md
Jonghakseo: - https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/613/commits/2762ac8f4238c5f2508a65500bde0f17dee18b52

😄

packages/i18n/README.md
Jonghakseo: Thx! I added example.

[feat: Add placeholder example](https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/613/commits/fa0a41b2f5fcd7497809745a91823ebec36b1317)

Jonghakseo:
@PatrykKuniczak I'm gonna merge this for now, and enahnce the documentation!

Issue #562, opened 2024-07-11
Discord Server


PatrykKuniczak:
I think good idea could be to open a server where we're(contributors) can talk more freely with each other on 1 channel.
And on other rest devs(which are only using our template) could sending a questions, like e.g Nest have.

@Jonghakseo What do you think about my idea, are you eager to open? 🥇

Jonghakseo:
Okay~! I'll try to make one sometime in July (before the beginning of August at the latest).

Jonghakseo:
https://discord.gg/Nu5Rf8xr

grolox69:
Hey @Jonghakseo, can you please update the invite? This link expired

Jonghakseo:
> Hey @Jonghakseo, can you please update the invite? This link expired

Sure

https://discord.gg/YZRG2z6Tsg

PR #616, opened 2024-07-27
Better firefox support

govza:
Firefox support improvement

## Priority*

## Purpose of the PR*

I was thinking to add e2e tests for both browsers. For Firefox running in Puppeter extension needs to be zipped and installed.
This PR allows zipping of Firefox extension and cleans Firefox related turbojs scripts

## Changes
feat: zipped firefox extension ready to work 
fix: zipper bundles on windows with posix slashes
refactor: zip command respects firefox settings
refactor: extension scripts depend on turbojs global firefox environment variable 

## How to check the feature

turbojs scripts refactoring:
(run `pnpm dev`) check Chrome can load extension (Firefox can't) [chrome://extensions/](chrome://extensions/)
(run `pnpm dev:firefox`) Firefox can load extension (Chrome can't)  [about:debugging#/runtime/this-firefox](about:debugging#/runtime/this-firefox)

zipper work:
(run `pnpm zip`) Chrome extension packed. Can be installed by drag-and-drop to [chrome://extensions/](chrome://extensions/)
(run `pnpm zip:firefox`) Firefox extension packed. Extension can be installed by "Load Temporary Add-on" to [about:debugging#/runtime/this-firefox](about:debugging#/runtime/this-firefox)

Also have to change turbojs dependsOn, since changing FIREFOX variable didn't cleaned old dist folder resulting in wrong zip bundle.

## Reference
globalEnv: Changes to the values of any environment variables in this list will change the hash for all tasks.
Firefox doesn't allow backslashes in the zip file. <https://github.com/mozilla/addons/issues/8879>

### DIFF
--- a/packages/dev-utils/lib/manifest-parser/impl.ts
+++ b/packages/dev-utils/lib/manifest-parser/impl.ts
@@ -25,6 +25,12 @@ function convertToFirefoxCompatibleManifest(manifest: Manifest) {
   manifestCopy.content_security_policy = {
     extension_pages: "script-src 'self'; object-src 'self'",
   };
+  manifestCopy.browser_specific_settings = {
+    gecko: {
+      id: 'example@example.com',
+      strict_min_version: '109.0',
+    },
+  };
   delete manifestCopy.options_page;
   return manifestCopy as Manifest;
 }
--- a/packages/zipper/lib/zip-bundle/index.ts
+++ b/packages/zipper/lib/zip-bundle/index.ts
@@ -1,5 +1,5 @@
 import { createReadStream, createWriteStream, existsSync, mkdirSync } from 'fs';
-import { resolve, relative } from 'path';
+import { resolve, posix } from 'path';
 import glob from 'fast-glob';
 import { AsyncZipDeflate, Zip } from 'fflate';
 
@@ -94,10 +94,13 @@ export const zipBundle = async (
       if (aborted) return;
 
       const absPath = resolve(distDirectory, file);
-      const relPath = relative(distDirectory, absPath); // Get the relative path
+      const absPosixPath = posix.resolve(distDirectory, file);
+      const relPosixPath = posix.relative(distDirectory, absPosixPath);
+
+      console.log(`Adding file: ${relPosixPath}`);
       streamFileToZip(
         absPath,
-        relPath, // Use relative path
+        relPosixPath, // Use relative path
         zip,
         () => {
           aborted = true;
--- a/packages/zipper/lib/zip-bundle/index.ts
+++ b/packages/zipper/lib/zip-bundle/index.ts
@@ -98,7 +98,7 @@ export const zipBundle = async (
       console.log(`Adding file: ${relPosixPath}`);
       streamFileToZip(
         absPath,
-        relPosixPath, // Use relative path
+        relPosixPath,
         zip,
         () => {
           aborted = true;
--- a/packages/zipper/lib/zip-bundle/index.ts
+++ b/packages/zipper/lib/zip-bundle/index.ts
@@ -45,17 +45,17 @@ export const zipBundle = async (
   {
     distDirectory,
     buildDirectory,
-    distDirectoryName,
+    archiveName,
   }: {
     distDirectory: string;
     buildDirectory: string;
-    distDirectoryName: string;
+    archiveName: string;
   },
   withMaps = false,
 ): Promise<void> => {
   ensureBuildDirectoryExists(buildDirectory);
 
-  const zipFilePath = resolve(buildDirectory, `${distDirectoryName}.zip`);
+  const zipFilePath = resolve(buildDirectory, archiveName);
   const output = createWriteStream(zipFilePath);
 
   const fileList = await glob(
--- a/packages/zipper/lib/zip-bundle/index.ts
+++ b/packages/zipper/lib/zip-bundle/index.ts
@@ -74,8 +74,6 @@ export const zipBundle = async (
     let totalSize = 0;
     const timer = Date.now();
     const zip = new Zip((err, data, final) => {
-      if (aborted) return;
-
       if (err) {
         pReject(err);
       } else {
--- a/turbo.json
+++ b/turbo.json
@@ -1,6 +1,7 @@
 {
   "$schema": "https://turbo.build/schema.json",
   "ui": "tui",
+  "globalEnv": ["__FIREFOX__"],
   "tasks": {
     "ready": {
       "outputs": [
@@ -16,18 +17,12 @@
         "dist/**",
         "build/**"
       ],
-      "env": [
-        "__FIREFOX__"
-      ],
       "persistent": true
     },
     "build": {
       "dependsOn": [
         "^build",
         "ready"
-      ],
-      "env": [
-        "__FIREFOX__"
       ],
       "outputs": [
         "../../dist/**",
@@ -37,7 +32,7 @@
     },
     "zip": {
       "dependsOn": [
-        "^build"
+        "build"
       ],
       "cache": false
     },
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "build": "turbo build",
     "build:firefox": "cross-env __FIREFOX__=true turbo build",
     "zip": "turbo zip",
+    "zip:firefox": "cross-env __FIREFOX__=true turbo zip",
     "dev-server": "pnpm -F hmr ready && pnpm -F hmr dev",
     "dev": "turbo ready && turbo watch dev --concurrency 20",
     "dev:firefox": "turbo ready && cross-env __FIREFOX__=true turbo watch dev --concurrency 20",
--- a/packages/zipper/index.ts
+++ b/packages/zipper/index.ts
@@ -6,5 +6,5 @@ import { zipBundle } from './lib/zip-bundle';
 zipBundle({
   distDirectory: resolve(__dirname, '../../dist'),
   buildDirectory: resolve(__dirname, '../../dist-zip'),
-  distDirectoryName: 'extension',
+  archiveName: process.env.__FIREFOX__ ? 'extension.xpi' : 'extension.zip',
 });
--- a/pages/content-runtime/lib/root.tsx
+++ b/pages/content-runtime/lib/root.tsx
@@ -15,25 +15,23 @@ export function mount() {
 
   const shadowRoot = root.attachShadow({ mode: 'open' });
 
-  /** Inject styles into shadow dom */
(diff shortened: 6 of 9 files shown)


PatrykKuniczak (changes requested):
Look good, but you don't change it everywhere(in all packages) and in root package.

And it's not possible to load this to firefox

image

After this edit, we can test it 😄

govza:
> And it's not possible to load this to firefox
> 
Sorry I cannot reproduce it. And pretty new to turborepo, maybe some flaky staff but it works for me... can you add steps and env to check?

> Look good, but you don't change it everywhere(in all packages) and in root package.
> 

Only build needs it from what I see and even zipper doesn't care, maybe I should remove... Since it's template I'm more pro minimal setup.

Also most packages doesn't have node at runtime and don't care about FF on build time, but if you have ideas, I'm open to discussion. BTW do you have discord or smth? 😃

PatrykKuniczak:
@govza We waiting for discord #562 i will happened soon.

As you can see __FIREFOX__ is loaded as a global env.

And it was using for build specific version for firefox, if i want to build it only with 'pnpm dev' or 'pnpm build' it will work only for chrome.

Your changes suggesting to drop ":firefox" versions from packages, and create only 1 version with 
```json
"build": "tsc --noEmit && cross-env __FIREFOX__=$__FIREFOX__ && vite build",
```
Look for example to 'manifest.js' and see difference between versions.

After all, you change this script in chrome-extension package, but this is folder with extension specific configs and additionally background script.

You don't change it on root and the turborepo works like:
```
You run 'pnpm dev' script on root, i will do it for each package and everything will run at one time.
```

If you want to create workaround/sugar for our solution, you need to rebuild all packages and create some magic, because IDN how to create it to work with 1 command, because of difference between output 'manifest.json'

It's highly possible, this isn't doable because of this case which i described.

If you want to change it only for zipper, this isn't good idea, because of. If we want to change sth in the future, to adjust this package for firefox and you removed it now, we need to rollback this PR to back to life this support.

IMO it isn't good idea to remove it from several packages because of the small benefit for now.

If you can explain, why this changes in ```impl.ts``` and ```zip-bundle/index.ts``` is necessary, and what's the prons of this solution?

turbo.json
PatrykKuniczak, line 1: Changes in this file is maybe good, but @Jonghakseo need to take a look, because he implementing this turbo, and he knows, why did he do it here, while he was implementing this 😄

turbo.json
```diff
@@ -37,7 +32,7 @@
     },
     "zip": {
       "dependsOn": [
-        "^build"
+        "build"
```
PatrykKuniczak, line 35: Why you remove it from here, but in the above ```build``` it's still?
Where's the difference, it's between configs in this package and the rest?
I wasn't configuring turbo before, and i don't know if it's really necessary 😃

govza:
Thank you for your clarification! 

I reviewed again the behavior of turbo and it looks like
  "globalEnv": ["__FIREFOX__"], actually sets up __FIREFOX__ environment variable for all the packages, so there is no need to redefine it in each level.

I guess it resolves your concerns mentioned that one may want to customize further each package depending on env variable. 

Should assume though that package scripts (at least related to FF) should be called only from the main package scripts (via turbo tasks). I don't see why one will call them separately but maybe there is use case.

> If you can explain, why this changes in `impl.ts` and `zip-bundle/index.ts` is necessary, and what's the prons of this solution?

change in impl.ts allows create working Firefox test bundle which can be used for e2e testing
change in zip-bundle/index.ts allows create working Firefox bundle on windows machine.
See Reference part for more info.

Hope it explains a bit the changes.

turbo.json
govza: Actually this is pure experimental finding, to make `zip:firefox` and `zip` properly build corresponding dist (with proper manifest).

It was not working with ^build, I can only assume that it doesn't send down environment variables to the children build scripts.

packages/dev-utils/lib/manifest-parser/impl.ts
```diff
@@ -25,6 +25,12 @@ function convertToFirefoxCompatibleManifest(manifest: Manifest) {
   manifestCopy.content_security_policy = {
     extension_pages: "script-src 'self'; object-src 'self'",
   };
+  manifestCopy.browser_specific_settings = {
+    gecko: {
+      id: 'example@example.com',
+      strict_min_version: '109.0',
```
govza, line 31: manifest v3 support enabled from version FF 109.0
id, maybe should come from env, or some other field of manifest.json.
At the end of the day, it needs valid email here to make it work

PatrykKuniczak:
@govza We running scripts from root, also zipper, because it's easier, all other scripts are starting from root, and there's no need to complicate it, to:
User need to go to subdirectory, for create a zip and then go back, to root to use other commands.

For this package we don't need to run all modules, as you described, but it maybe could be helpful in the future to have FIREFOX env declared with dedicated scripts in package.

Maybe your changes is good, because of course for now we don't using __FIREFOX__ env, and maybe it's never necessary to use and we can throw it away, but i'm not sure about that, let's wait to @Jonghakseo review, he know well about entire project and have better knowledge about this infrastructure :)

About ```zip-bundle/index.ts``` i was testing it, and it's works for my on windows, now i test it one more time, and still working.
But YEAH you're right, i didn't test it on firefox, there's a lack of zipper for firefox, good to you ❤️ 

This script is indeed necessary.
Now it's working for FF also.

But there's some issue, because in the zipper folder, this 'undefined' folder is creating 
image

And nested temp and then inside with 'tsx-my PC name' and i think it could be related to this 'posix' because ealier i don't recognize it.

Last think/Request is.
Lets rollback changes in ```chrome-extension/package.json``` and we will take a look on that in other PR, because it could be done better, i can refactor it tomorrow.

Jonghakseo (commented):
I like the changes. I'll have to do some more testing... I'm a bit skeptical about the support for Firefox, it's a lot different from calling the API in the Chrome browser in a lot of small ways than I thought. It will be hard for us to give the same development experience on Firefox as the Chrome... but it's worth the effort!

govza:
@PatrykKuniczak You agreed that we run scripts from root, so basically there is no need to redefine environment variables in the children. Isn't it cleaner this way why put them back? And tomorrow never come, I guess. It's always like that :( 
I think zip temp folders may occur due to script interruption or smth. Let's see if the problem persists.

@Jonghakseo indeed, even testing with Firefox is much harder. There is some work with unified Bidi protocol, I was thinking to experiment with it.

Anyway, please feel free request upgrades or whatever, happy to participate.1

PatrykKuniczak:
> @PatrykKuniczak You agreed that we run scripts from root, so basically there is no need to redefine environment variables in the children. Isn't it cleaner this way why put them back? And tomorrow never come, I guess. It's always like that :( 
> I think zip temp folders may occur due to script interruption or smth. Let's see if the problem persists.
> 
> @Jonghakseo indeed, even testing with Firefox is much harder. There is some work with unified Bidi protocol, I was thinking to experiment with it.
> 
> Anyway, please feel free request upgrades or whatever, happy to participate.1

Yeah vars are redeclared i think unmecessary, tomorrow comes but i want more time, cause i don't have enough time yesterday, because of busy time at life.

I want to refactor it completetly and do sth like ```base.package.script``` and inherit it in all packages, to simplify it and avoid redundancy.

In open source projects it look completely different than in company projects, it goes so much slower, but it's ok, it's non profil side project, be pacient 🙂

chrome-extension/package.json
PatrykKuniczak, line 1: Rollback this changes we will take a look on that on #625

PatrykKuniczak (changes requested):

PatrykKuniczak (commented):
Works perfect for my.

But i was thinking, to add 
https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/issues/641

In this PR
What are you guys thinking about this idea?

govza:
@PatrykKuniczak I have another one with e2e tests POC if we merge this one already.

PatrykKuniczak:
> @PatrykKuniczak I have another one with e2e tests POC if we merge this one already.

Oo, then you want to merge it in this shape, yeah?

PatrykKuniczak (approved):
LGTM

PatrykKuniczak:
@govza 1 more think

image

Why this ```undefined``` tree are growing on ```pnpm zip``` and ```pnpm zip:firefox```

PatrykKuniczak (changes requested):
wait, till ```undefined``` folder will be explained

govza:
> wait, till `undefined` folder will be explained

I cannot reproduce it. Can you reproduce it and give some details about OS and node version?

PatrykKuniczak:
@govza Win 11, Node: 20.11.1

PatrykKuniczak:
@govza Now i'm checking your last changes about styles, after this i can try to figure it out.

PatrykKuniczak:
@govza There's possibility to aborted be different than ```false``` here?
```ts
    const zip = new Zip((err, data, final) => {
      if (aborted) return;
```

PatrykKuniczak:
@govza Ok, this was some kind of conflict, when i was changing branches, i need to remove node_modules manually and reinstall, and it works.

Give me response to comment above and we are on the home stretch 😄

govza:
> @govza There's possibility to `aborted` be different than `false` here?
> 
> ```ts
>     const zip = new Zip((err, data, final) => {
>       if (aborted) return;
> ```
No, I don't think so.

PatrykKuniczak (approved):
Second LGTM 😆

Jonghakseo (approved):
LGTM

packages/zipper/lib/zip-bundle/index.ts
```diff
@@ -94,10 +92,13 @@ export const zipBundle = async (
...
       const absPath = resolve(distDirectory, file);
-      const relPath = relative(distDirectory, absPath); // Get the relative path
+      const absPosixPath = posix.resolve(distDirectory, file);
+      const relPosixPath = posix.relative(distDirectory, absPosixPath);
+
+      console.log(`Adding file: ${relPosixPath}`);
       streamFileToZip(
         absPath,
-        relPath, // Use relative path
+        relPosixPath, // Use relative path
```
PatrykKuniczak, line 101:
```suggestion
        relPosixPath,
```

@govza This is your TODO, i think :)

PatrykKuniczak: > It's not to-do but a comment, not mine
> 
> I'm leaving for holidays, so feel free to do what's necessary and merge 😉

I will do, next week.

Have a good time on holiday ❤

PR #640, opened 2024-08-06
fix/isFirefox const to be boolean

PatrykKuniczak:

## Priority*

## Purpose of the PR*
There's missing check, and because of that the value of const, isn't boolean

## Changes*
add equation


Jonghakseo (approved):
LGTM

PR #650, opened 2024-08-08
fix/not permitted unlink for 'pnpm clean'

PatrykKuniczak:

## Priority*

## Purpose of the PR*
Give users possibility to use ```pnpm clean``` script in 100% :)

## Changes*
Add shut down turbo daemon before rimraf 

## How to check the feature
run ```pnpm clean```

## Reference
I was thinking also about total shut down of daemon, but i'm not sure, we're really need it.
@Jonghakseo If we don't need it, we can turn it off globally, but it isn't working properly for now.
I was sent question to turborepo community, let's wait for response :)
https://github.com/vercel/turbo/pull/8728#issuecomment-2275534895

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "url": "https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite.git"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


Jonghakseo (approved):
LGTM

PR #627, opened 2024-08-03
Fixed syntax error in pages/content-ui/package.json

codergigachad:
> `*` Please fill in the required items.

## Priority*

- [x] High: This PR needs to be merged first for other tasks.
- [ ] Middle: This PR should be merged quickly to prevent conflicts due to common changes. (default)
- [ ] Low: This PR does not affect other tasks, so it can be merged later.

## Purpose of the PR*
Tailwind build was not working in content-ui during hmr due to syntax error.

## Changes*
Changed the package.json file of pages/content-ui

PR #608, opened 2024-07-26
Add Hook to Compile Tailwind-Output on PMPM Install

Leizhenpeng:
To prevent potential issues during the first compilation when the tailwind-output file is ignored, a hook should be added to ensure that this file is compiled during pmpm install. This will guarantee the necessary files are prepared for the user's environment.

relate
--fix #600 #601 #596

PR #654, opened 2024-08-13
fix/tw-classnames-output-#627

PatrykKuniczak:

## Priority*

## Purpose of the PR*
I can't open PR to fork, that's why i created this PR.

This is related to #627.

I fixed that fix :)

Solution from #627 wasn't good enough because of the never ending build and it's still don't work properly.

## Changes*

I added ```concurrently``` for run both watch scripts on the same time, now it was running independly.

Without that, only first script works.
If i placed vite first then tailwind output file wasn't regenerated.
Else Vite don't run and HMR don't refresh page.

I added also ```-m``` flag, for minify output file.

And i added ```dist``` folder inside ```content-ui```

## How to check the feature
Edit files from ```content-ui``` page.


codergigachad:
My solution was a beginner solution, thanks for fixing it like an expert.

PatrykKuniczak:
> My solution was a beginner solution, thanks for fixing it like an expert.

Thanks, but i don't feel like an expert.

You've found that issue, which we're not aware, that's also very helpful 😄

pages/content-ui/package.json
```diff
@@ -10,15 +10,14 @@
   ],
   "scripts": {
     "clean": "rimraf ./dist",
-    "build:tailwindcss": "tailwindcss -i src/tailwind-input.css -o src/tailwind-output.css",
-    "build": "pnpm run clean && pnpm type-check && pnpm build:tailwindcss && vite build",
-    "build:watch": "cross-env __DEV__=true vite build --mode development & pnpm build:tailwindcss -- --watch",
+    "build:tailwindcss": "pnpm dlx tailwindcss -i ./src/tailwind-input.css -o ./dist/tailwind-output.css -m",
```
Jonghakseo, line 13: 스크린샷 2024-08-14 오후 8 45 29

I saw above error message. How about moving it inside `src`, the original file's location, instead of the `dist` folder?

Is there a reason you moved the output file to dist?

```suggestion
    "build:tailwindcss": "pnpm dlx tailwindcss -i ./src/tailwind-input.css -o ./src/tailwind-output.css -m",
```

pages/content-ui/src/index.tsx
```diff
@@ -1,8 +1,7 @@
 import { createRoot } from 'react-dom/client';
 import App from '@src/app';
-// eslint-disable-next-line
-// @ts-ignore
-import tailwindcssOutput from '@src/tailwind-output.css?inline';
+// @ts-expect-error That's because output file is create during build
+import tailwindcssOutput from '../dist/tailwind-output.css?inline';
```
Jonghakseo, line 4:
```suggestion
import tailwindcssOutput from '@src/tailwind-output.css?inline';
```

This is related to the comment above.

Jonghakseo (commented):
😄

pages/content-ui/package.json
PatrykKuniczak: @Jonghakseo I want to separate output file from src, and there's entire /dist ignored.

Jonghakseo: - https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/608

Like the PR above, It seems like a good idea to exclude from VSC for output files.

PatrykKuniczak: @Jonghakseo What do you mean about 'out of vsc'?

PatrykKuniczak: That bug occur probably because of the lack of '&'
In `build-watch` script:
`development & pnpm build:tailwindcss`

Not it's done by `concurrently`

And i check it a minute ago on VSC, and works perfect, `postinstall` isn't necessary.

codergigachad:
@PatrykKuniczak any progress here? cause this is very important.

pages/content-ui/package.json
Jonghakseo: Sorry. It was too short. I mean Version Control System.

> @Jonghakseo What do you mean about 'out of vsc'?

Jonghakseo: 스크린샷 2024-08-18 오후 2 18 38

For me, I still see the above error message.
I just run `pnpm dev` after `pnpm clean` and I can see it right away.

PatrykKuniczak:
> @PatrykKuniczak any progress here? cause this is very important.

We're waiting for @Jonghakseo review

PatrykKuniczak:
@codergigachad OO maybe he done review, before my message, i'm on mobile then and don't see all messages :)

I will take a look to merge it today.

pages/content-ui/package.json
PatrykKuniczak: @Jonghakseo Yeah, that's reproducible, i need to fix it.

PatrykKuniczak: @Jonghakseo Ok, I added a tw prebuild before, `postinstall` can't solve that issue, but this kind of prebuild did it.

Jonghakseo: I'm gonna check it!

Jonghakseo: HMR was not working. so i fixed it. 

[fix: use dev env](https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/654/commits/3f3bd09ba9f3fd98537f325857fde3a8ac49a50b)

Jonghakseo (approved):

PatrykKuniczak:
@Jonghakseo I belived i had to check, because i was thinking if this env will be resolved properly.

Good to you, thank for the second check ❤️

PR #658, opened 2024-08-15
feat/create utils package

PatrykKuniczak:

## Priority*

## Purpose of the PR*
I want to create utils, like dev-utils which are already exists.

Users will have already predefined module for save them utils functions and types like this example "ValueOf"

## Changes*

Create 'utils' package and move "ValueOf" from 'dev-utils' to 'utils'


packages/utils/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "@extension/utils",
+  "version": "0.2.2",
```
Jonghakseo, line 3:
```suggestion
  "version": "0.3.0",
```

packages/utils/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "main": "dist/index.js",
+  "module": "dist/index.js",
+  "types": "index.ts",
+  "scripts": {
+    "clean": "rimraf ./dist && rimraf ./build",
+    "ready": "tsc",
+    "lint": "eslint . --ext .ts,.tsx",
+    "lint:fix": "pnpm lint --fix",
+    "prettier": "prettier . --write --ignore-path ../../.prettierignore",
+    "type-check": "tsc --noEmit"
```
Jonghakseo, line 19: If we add functions in the future, we'll need to set esbuild or something... How about adding a function example?

Jonghakseo (commented):
The definitions of packages inside monorepo seem to be getting a bit fuzzy and need to be clear.

There also seems to be some overlap with `shared`.

PatrykKuniczak:
@Jonghakseo Maybe put this `utils` inside `shared` next to `hooks` ?

Jonghakseo (approved):
LGTM

PR #659, opened 2024-08-15
chore/simplify storage

PatrykKuniczak:

## Priority*

## Purpose of the PR*
I want to create this base.ts file, a little smaller and easier to read for future :)

## Changes*
I renamed get func, because this _private isn't necessary here IMO, because it's exported in return afterall.

And i removed some unnecessary thinks and create some space between.


Jonghakseo (approved):
LGTM

PR #664, opened 2024-08-17
fix/handle error on injecting script from popup

PatrykKuniczak:

## Priority*

## Purpose of the PR*
I want to handle that, because error was going to extension console, which our users don't like.

Also devs on dev process, can forget/don't know about that error and users could be shocked why script isn't working here :)

## Changes*
I done 
an simple alert notification to let users know what's wrong.

It's more meaningful than before error.
That's simply but it's sufficient IMO, because each dev can handle that in their way

## How to check the feature
Try to inject script from popup in place where it's not possible, e.g chrome home page


Jonghakseo (approved):
LGTM

PR #667, opened 2024-08-18
[Enhance] Add messaging example

Jonghakseo:

## Priority*

## Purpose of the PR*

Adds a simple implementation for easy use of the Messages API.

## Changes*

- Add the messaging package
- Separate ready from prepare, a ready command that runs only once

## How to check the feature

Use messaging api :)


Jonghakseo:
# README.md preview

---
# Messaging Package

This package contains code messaging api. 
To use the code in the package, you need to add the following to the package.json file.

```json
{
  "dependencies": {
    "@extension/messaging": "workspace:*"
  }
}
```

## Messaging API

You can use messaging api to send/receive messages between context.

### Usage

```typescript
import { messaging } from '@extension/messaging';

// Send message
messaging.send('message', { data: 'data' });

// Receive message
messaging.on('message', (payload) => {
  console.log(payload.data); // 'data'
  return 'response';
});
```

### Message Types

You should define message types in the `type.ts` file.

```typescript
export interface PayloadAndResponse<P = unknown, D = unknown> {
  payload: P;
  response: D;
}
/**
 * Define the type of messages
 * If you want to add a new message type, you can add it here.
 */
export interface Messages {
  greeting: PayloadAndResponse<string, string>;
  //...
}
```

Then you can take advantage of the message types in the `messaging` api.

```typescript
import { messaging } from '@extension/messaging';


messaging.send('greeting', "Hello").then((response) => {
  console.log(response); // Hi
});
```


```typescript
import { messaging } from '@extension/messaging';

messaging.on('greeting', (payload) => {
  console.log(payload); // Hello
  return 'Hi'; // response
});
```

### SendToCurrentTab

You can use `sendToCurrentTab` to send a message to the current tab.

```typescript
import { messaging } from '@extension/messaging';

messaging.sendToCurrentTab('message', "this is current tab!");
```

You can receive the message in the current tabs `content script`. (e.g. `content` or `content-ui`) 

```typescript
import { messaging } from '@extension/messaging';

messaging.on('message', (payload) => {
  // Do something
});
```

---

packages/messaging/messaging.ts
```diff
@@ -0,0 +1,171 @@
...
+
+const ERROR_SUFFIX = '__Error';
+
+function sendByPort<T extends MessageType>(type: T, payload?: Payload<T>): Promise<Response<T>> {
+  return new Promise((resolve, reject) => {
+    const port = chrome.runtime.connect();
+
+    port.onMessage.addListener((message: Message<T>) => {
+      const type = message.type;
+      const response = message.response as Response<T>;
```
PatrykKuniczak, line 24:
```suggestion
    port.onMessage.addListener(({ type, response }: Message<T>) => {
```
And type for response should be taken from Message

packages/messaging/messaging.ts
```diff
@@ -0,0 +1,171 @@
...
+const send = <T extends MessageType>(type: T, payload?: Payload<T>, options?: SendOptions): Promise<Response<T>> => {
+  const { by = 'port' } = options ?? {};
+
+  if (by === 'sendMessage') {
+    return sendBySendMessage<T>(type, payload);
+  }
+  return sendByPort<T>(type, payload);
+};
+
+const createPortUtils = (sendMessage: (message: unknown) => unknown) => {
```
PatrykKuniczak, line 82: message should take generic type

PatrykKuniczak (changes requested):
I think adding this to all pages `package.json` isn't necessary, because users may never used this, and it's additional deps to install.

Rather place it inside `README`

And one more thing refactor from `ready` to `prepare` should be in another PR, but it'll be dumb to move it now, let's leave it there xD

Rest on Discord

chrome-extension/src/background/index.ts
```diff
@@ -7,3 +8,7 @@ exampleThemeStorage.get().then(theme => {
 
 console.log('background loaded');
 console.log("Edit 'chrome-extension/src/background/index.ts' and save to reload.");
+
+messaging.on('foo', (payload, sender) => {
+  return `background received ${payload} message from ${sender?.url}`;
```
govza, line 13: background received [object Object] message from https://example.com/

packages/messaging/type.ts
```diff
@@ -0,0 +1,11 @@
+export interface PayloadAndResponse<P = unknown, D = unknown> {
+  payload: P;
+  response: D;
+}
+/**
+ * Define the type of messages
+ * If you want to add a new message type, you can add it here.
+ */
+export interface Messages {
+  foo: PayloadAndResponse<{ bar: string }, string>;
```
govza, line 10: Maybe more meaningful naming, it's pretty hard to understand at first with foo: bar, what is the protocol what is request types and return types. Maybe HandshakeProtocol, RequestHandshake, ResponseHandshake etc.?

packages/messaging/messaging.ts
```diff
@@ -0,0 +1,171 @@
...
+
+export const messaging = {
+  sendToCurrentTab,
+  send,
+  sendByPort,
+  sendBySendMessage,
+  on,
+  onByPort: addHandlerByPort,
+  onBySendMessage: addHandlerBySendMessage,
+};
```
govza, line 171: Some documentation / jsdoc when / how to use? From background, from content-script send etc?

packages/messaging/type.ts
govza, line 1: Shouldn't it go to the packages/shared/utils/shared-types?

Jonghakseo:
- https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/747

This might be a better alternative, I'll look into it.

PR #697, opened 2024-09-03
chore/refactor-all-packages-json

PatrykKuniczak:

## Priority*

## Purpose of the PR*
I want to do it in more `turbo` friendly way, and i've done it as simply as possible for now.

## Changes*
I've split `clean` command and remove redundancy from `dev`  

I've used `pnpx` for avoid cases when users run `pnpm clean:turbo` and then want to run `pnpm clean:node_modules` or `clean`

By this move i want to avoid a bunch of issues about new error(users reports) 😆 

I've also removed tasks from `dependsOn` due to lack of opportunities to define a order, in `package.json` i can do it freely.
Maybe in the future we'll find a better solution for that. 

And also don't pay attension to `cross-env` because it works for now, and i'll try to refactor it, to looks and works better 😄 

## How to check the feature
Run new and old commands, and check if it works


packages/i18n/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@extension/i18n",
   "version": "0.3.1",
-  "description": "chrome extension internationalization",
+  "description": "chrome extension - internationalization",
```
Jonghakseo, line 4: 👍🏼

package.json
```diff
@@ -7,19 +7,23 @@
...
-    "clean:turbo": "turbo clean && turbo daemon stop && rimraf dist && rimraf .turbo",
-    "clean": "pnpm clean:turbo && pnpm clean:node_modules",
+    "clean-root:bundle": "pnpx rimraf dist && pnpx rimraf dist-zip",
+    "clean-root:node_modules": "pnpx turbo daemon stop && pnpx rimraf node_modules",
+    "clean-root:turbo": "pnpx turbo daemon stop && pnpx rimraf .turbo",
+    "clean-root": "pnpm clean-root:bundle && pnpm clean-root:node_modules && pnpm clean-root:turbo",
+    "clean:node_modules": "pnpx turbo clean:node_modules && pnpm clean-root:node_modules",
+    "clean:turbo": "pnpx turbo clean:turbo && pnpm clean-root:turbo",
+    "clean:bundle": "pnpx turbo clean:bundle && pnpm clean-root:bundle",
+    "clean": "pnpx turbo clean && pnpm clean-root",
```
Jonghakseo, line 19: OK... There's too many clean scripts lol

I think in many case we need this clean scripts just for clean entire projects stuff. So I suggest leaving just the clean-root scripts.

```suggestion
    "clean:bundle": "pnpx rimraf dist && pnpx rimraf dist-zip",
    "clean:node_modules": "pnpx turbo daemon stop && pnpx rimraf node_modules",
    "clean:turbo": "pnpx turbo daemon stop && pnpx rimraf .turbo",
    "clean": "pnpm clean-root:bundle && pnpm clean-root:node_modules && pnpm clean-root:turbo",
```

Jonghakseo (commented):
👍🏼

chrome-extension/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": "@extension/tsconfig/app",
   "compilerOptions": {
-    "baseUrl": "./",
+    "baseUrl": ".",
```
Jonghakseo, line 4: Was this change necessary?

PatrykKuniczak: No, but i want to do it, to have this config like all others

package.json
PatrykKuniczak: That suggestion isn't fully good, because of the mistake in `clean` but it's good change to make, i've done it.

I need to run turbo for the rest `clean:` scripts, check it and feel free to request more changes if needs :)

Jonghakseo (approved):
LGTM

PR #704, opened 2024-09-06
fix: vite config types error

PatrykKuniczak:

## Priority*

## Purpose of the PR*
I need to clean up my own mess :)
There's need to unify versions between packages, to have no errors on `vite.config.mts` of `chrome-extension` package

## Changes*
I've unified versions of `root` and `chrome-extension` package.json

## How to check the feature
Run e.g `pnpm lint`, and check it runs


chrome-extension/package.json
```diff
@@ -31,6 +31,7 @@
     "ts-loader": "^9.5.1",
     "deepmerge": "^4.3.1",
     "cross-env": "^7.0.3",
-    "@types/node": "^20.16.3"
+    "@types/node": "^20.16.5",
+    "vite": "^5.4.3"
```
Jonghakseo, line 35: It's already installed as root, so I guess I can uninstall it!

```suggestion
    "@types/node": "^20.16.5"
```

PatrykKuniczak: Yeah, you're right.
I want to do it, but i've forgot, because i have problem with this lack of `turbo ready` for `dev` and it doesn't working.
It working perfect without it, because it's redundancy 😄

Jonghakseo (approved):
LGTM

PR #725, opened 2024-09-10
feat/create better solution for env vars

PatrykKuniczak:

## Priority*

## Purpose of the PR*
I've decided to fix issue about `import.meta.env` by solution with only `process.env` because i hope this will be more simple, to have only one instance with all envs

## Changes*
I've removed vite env solution and replace it with `process.env`

## How to check the feature
Try to add env(Read docs about it) and try to access it anywhere


Jonghakseo (approved):
LGTM

Issue #714, opened 2024-09-09
Taking packages from root


PatrykKuniczak:
**Is your feature request related to a problem? Please describe.**
We need to install package when we want to use it inside some module, but in the same time this package are installed on root `package.json`

**Describe the solution you'd like**
I think it could be good idea to install packages which are used by min. 2 packages in root `package.json` and then inherit it inside specific module `tsconfig.json` under the `includes` key.

PR #727, opened 2024-09-10
fix: types for `chrome-extension` package to be compatible

PatrykKuniczak:

## Priority*

## Purpose of the PR*
I've described it in details here: https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/issues/714#issuecomment-2341656789

## Changes*
I've added `vite` and `@types/node` for `chrome-extension` package

### DIFF
--- a/chrome-extension/package.json
+++ b/chrome-extension/package.json
@@ -30,6 +30,8 @@
     "magic-string": "^0.30.10",
     "ts-loader": "^9.5.1",
     "deepmerge": "^4.3.1",
-    "cross-env": "^7.0.3"
+    "cross-env": "^7.0.3",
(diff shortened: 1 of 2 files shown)


Jonghakseo (commented):
Someone might question why we need to install duplicate packages again 🤔 

How about adding them to peerDependencies instead? I'm sure there's a better solution, hmmm...

First, let's merge the PR to fix the problem. We'll put in a PR to fix the issue first, and then replace it when we have a better solution.

Jonghakseo:
- https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/728

PR #731, opened 2024-09-12
chore: set off daemon for entire app

PatrykKuniczak:

## Priority*

## Purpose of the PR*
I've wanted to resolve issue with deleting `.turbo` and this resolve it and also improve performance, because of:
https://github.com/vercel/turborepo/issues/8122
On some machines

## Changes*
I've added `"daemon": true` to `turbo.json` and remove daemon stop from clean scripts

## How to check the feature
run e.g. `pnpm dev`, then stop script and try to run `pnpm clean:turbo` and check it works without any permission error


turbo.json
```diff
@@ -2,6 +2,7 @@
   "$schema": "https://turbo.build/schema.json",
   "ui": "tui",
   "globalEnv": ["__FIREFOX__"],
+  "daemon": false,
```
Jonghakseo, line 5: Are there any performance issues? I'm curious about the tradeoffs of disabling the daemon.

PatrykKuniczak: I didn't have any, but the main reason i've did it is, issue with no possibility of removal `.turbo` on Windows, with this i don't need to stop it from `package.json` script

PatrykKuniczak:
@Jonghakseo What's with this?

PR #743, opened 2024-09-15
refactor: theme storage improve usability

govza:
Separate user specific stores logic from generics

## Priority*

## Purpose of the PR*
Ease of sync of the forks with different types of storage

## Changes*
Moved storage implementations to separated folder

## How to check the feature
Check regression on storage, popup toggle theme also toggles options page theme

## Reference
N/A


packages/storage/lib/base/base.ts
PatrykKuniczak, line 1: Name should be `index.ts` like in other modules IMO

packages/storage/lib/impl/exampleThemeStorage.ts
```diff
@@ -1,6 +1,12 @@
...
-import type { Theme, ThemeStorage } from './types';
+import { StorageEnum } from '../base/enums';
+import { createStorage } from '../base/base';
+import type { BaseStorage } from '../base/types';
+
+type Theme = 'light' | 'dark';
+
+type ThemeStorage = BaseStorage<Theme> & {
+  toggle: () => Promise<void>;
+};
```
PatrykKuniczak, line 9: Why are you moved it from `types.ts`?
That file was created for store all kind of types from `storage`

PatrykKuniczak (changes requested):

packages/storage/lib/impl/exampleThemeStorage.ts
govza: Because It's related to user specific storage model and will be hassle to rebase or merge upstream to fork if smth changes on storage generics.

PatrykKuniczak: ok, that make sense

Jonghakseo (approved):
LGTM

packages/storage/lib/base/base.ts
govza: It's not file with exports of the directory modules.
I thought to not create such index, since these are not meant to be used outside of impl folder, except type which referenced explicitly.

PatrykKuniczak: But this is main file of this structure, IMO it should be index like in other modules

govza: From what I see all modules follow this example:

index.ts - list of exports from the folder
if it's only the file in folder than it is containing exports itself.

can you link the other modules example you referring to?

PatrykKuniczak: For example:
https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/blob/main/packages/zipper/lib/zip-bundle/index.ts

govza: >  if it's only the file in folder than it is containing exports itself.

PatrykKuniczak: OK right.

In other words.
Each folder of module should contain `index.ts` as a container file, which exporting all necessary staff from inside.

If you want to leave that file like it is ok, but let's create index file and export from `base` folder necessary things :)

PatrykKuniczak (approved):
LGTM

PR #711, opened 2024-09-08
feat/add full ts support/coverage

PatrykKuniczak:
> `*` Please fill in the required items.

## Priority*

- [x] High: This PR needs to be merged first for other tasks.
- [ ] Middle: This PR should be merged quickly to prevent conflicts due to common changes. (default)
- [ ] Low: This PR does not affect other tasks, so it can be merged later.

## Purpose of the PR*
As we was talking on Discord, i've decided to refactor project a little and create it more `ts` friendly. 
It'll be also very helpful for me, when i'll be working on custom env module, because i've some conflicts, now it should works perfect :)

Don't pay attention to redundant `cross-env` and overall env solutions because it'll be refactored #722

## Changes*
I've changed `.js` to `.ts` where it's necessary

## How to check the feature
Run it as normal `pnpm build` or/and `pnpm dev`
And check if all pages works well on Chrome and Firefox :)

PR #748, opened 2024-09-17
feat: bump node to 22

PatrykKuniczak:

## Priority*

## Purpose of the PR*
I've bump it, because we're planning to implement native WebSockets and i need to bump to at least to 20 for #711 works

## Changes*
I've bumped it to latest and enforce users to have that version for use project.

## How to check the feature
Nothing changed, let's try to use `pnpm build` or sth to check


govza:
It's entering LTS this October so should be fine.

But I'd keep most conservative one, like 20 LTS, just in case developer uses packages requiring older node.

Jonghakseo (approved):
LGTM

PR #749, opened 2024-09-17
added multi entryfiles for content-ui + internalize tailwindcss bundle

tngflx:
Feat : to add multi entryfiles for content-ui as ui page have lots of pages

## Priority*

## Purpose of the PR*
as chrome extension are not limited to only one subdomain, different part of code might be needed on different subdomains, its always great to be able to put your codes in boxes, and make it maintainable

## Changes*
Changes to vite.config.mts, and added a tailwind-bundler.ts

## How to check the feature

image


PatrykKuniczak:
@tngflx You don't want to work more on it?

PR #750, opened 2024-09-18
Streamlining production of tailwindcss bundled css, introducing multiple entryfiles for content-ui foldesr

tngflx:
**Make tailwindcss centralized css production settings, without creating postcss config in every package.json**
**Added multiple entryfiles just for content-ui**

## Priority*

## Purpose of the PR*
Streamlining css production for shadowdom, which included minification and treeshaking to be used for shadowdom
Added multiple entryfiles just for content-ui

## Changes*
1) Tailwindcss bundle production is a mess, i created a plugin to be shared accross all pages folder, without the need of redundant @tailwind base ... etc, to be pasted all over the code, the plugin itself contains the postcss config, and thereby internalizing the css production for shadowdom. And hence the command https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/blob/a7d51c3266803703e211223e49de675d119fdfa6/pages/content-ui/package.json#L15C1-L16C1 can be removed,
and this also fixes hmr crashes upon live editing of tailwindcss classess

2) Multi entry files for content-ui folder, as a normal dev will have multiple subdomains to inject across, its smarter to have multiple entry files injected for different subdomains, I've made the script to autogenerate folders and files inside content-ui, but u have to pay attention as well to make changes in manifest.js

## How to check the feature
Everything works even without the postcss config in package.json, and every pages now have its own generated global-out.css

just fork my repo and install as usual, you'll see all the tailwindcss still works perfectly

https://discord.com/channels/1263404974830915637/1264514968288759889/1285558712941023345
as you see above hmr doesn't crash

image
All the css is still working but with leaner size


pages/content-ui/script/entryPoints.ts
```diff
@@ -0,0 +1,52 @@
...
+  const entryPoints: Record<string, string> = {};
+
+  const foldersToProcess =
+    options.include ||
+    readdirSync(srcDir, { withFileTypes: true })
+      .filter((item): item is Dirent => item.isDirectory())
+      .map(item => item.name);
+
+  foldersToProcess.forEach(folderName => {
+    const shouldExclude = options.exclude && options.exclude.includes(folderName);
```
PatrykKuniczak, line 17: Maybe sth like that:

```suggestion
    const shouldExclude = options.exclude?.includes(folderName);
```

pages/content-ui/script/entryPoints.ts
```diff
@@ -0,0 +1,52 @@
...
+const rootDir = resolve(__dirname, '..');
+const srcDir = resolve(rootDir, 'src');
+
+export const getEntryPoints = (options: { include?: string[]; exclude?: string[]; createRecursively?: boolean }) => {
+  const entryPoints: Record<string, string> = {};
+
+  const foldersToProcess =
+    options.include ||
+    readdirSync(srcDir, { withFileTypes: true })
+      .filter((item): item is Dirent => item.isDirectory())
```
PatrykKuniczak, line 13: Why not like that?

```suggestion
      .filter((item: : Dirent) => item.isDirectory())
```

I don't really know this kind of typing, that's interesting

pages/content-ui/script/entryPoints.ts
```diff
@@ -0,0 +1,52 @@
+import { readdirSync, existsSync, mkdirSync, writeFileSync, Dirent, readFileSync, renameSync } from 'fs';
```
PatrykKuniczak, line 1:
```suggestion
import { readdirSync, existsSync, mkdirSync, writeFileSync, Dirent, readFileSync, renameSync } from 'node:fs';
```

pages/content-ui/script/entryPoints.ts
PatrykKuniczak, line 1: This shouldn't work that way, it need to edit `manifest.ts` before build.
If user have 3 scripts for different matches, then those scripts need to be taken with matches(from e.g special `const` like, Metadata for `NextJS` or sth like that

This should reading all folders from e.g. `matches` folder which are placed inside `src` and this special const from inside e.g. simpy `MATCHES`

pages/content-ui/src/page2/page2.tsx
```diff
@@ -0,0 +1,23 @@
+import { useEffect } from 'react';
+import { Button } from '@extension/ui';
+import { useStorage } from '@extension/shared';
+import { exampleThemeStorage } from '@extension/storage';
+
+export default function App() {
+  const theme = useStorage(exampleThemeStorage);
+
+  useEffect(() => {
+    console.log('content ui loaded');
```
PatrykKuniczak, line 10: Shouldn't be the same as in first script, but anyway.
1 example is enough

pages/content-ui/vite.config.mts
```diff
@@ -1,24 +1,38 @@
...
+  plugins: [processCssPlugin()],
   publicDir: resolve(rootDir, 'public'),
   build: {
     lib: {
-      entry: resolve(srcDir, 'index.tsx'),
+      entry: entryPoints,
       name: 'contentUI',
-      formats: ['iife'],
-      fileName: 'index',
+      formats: ['es'],
```
PatrykKuniczak, line 33: Don't have it to `es` because it isn't working good, i've checked it XD

pages/content-ui/vite.config.mts
```diff
@@ -1,24 +1,38 @@
...
+      fileName: (format, entryName) =>
+        isDev ? `${entryName}/index_dev.js` : `${entryName}/index.js`
```
PatrykKuniczak, line 35: Why are you need to have 2 different files?

pages/new-tab/src/index.tsx
```diff
@@ -1,6 +1,6 @@
 import { createRoot } from 'react-dom/client';
 import '@src/index.css';
-import '@extension/ui/lib/global.css';
+import '../dist/global-out.css';
```
PatrykKuniczak, line 3: Maybe your plugin are able to add this import during build to `.js` bundled file?
We can avoid those imports in each file.

packages/hmr/lib/plugins/tailwindcss-bundler.ts
```diff
@@ -0,0 +1,42 @@
...
+import LightningCSS from 'lightningcss';
+
+const rootDir = path.resolve(__dirname, '../../../../');
+
+export function processCssPlugin(): Plugin {
+  return {
+    name: 'process-css',
+    async configResolved(config) {
+      const inputFile = path.resolve(rootDir, 'ui/lib/global.css');
+      const outputFile = path.resolve(config.root, 'dist/global-out.css');
```
PatrykKuniczak, line 17: It's really necessary to change name to `-out`?

PatrykKuniczak (changes requested):

pages/content-ui/vite.config.mts
```diff
@@ -1,24 +1,38 @@
...
+ * Here is the place to add new entry points, u can straight away add new folders here to include
+ * without creating folders in src dir manually.
+ * @param {string[]} options.include - Folders to include !!MODIFY ME!!
+ * @param {string[]} options.exclude - Folders to exclude !!MODIFY ME!!
+ * @param {boolean} options.createRecursively - Create folders recursively
+ * @returns {Record<string, string>}
+ */
+const entryPoints = getEntryPoints({ 
+  include: ['page1', 'page2'], createRecursively: true 
+});
```
Jonghakseo, line 19: It seems more like `makeEntryponits` than `getEntryPoints`, and shouldn't it be possible to run it as a separate script? I'm curious why it's in the vite config together.

pages/content-ui/script/entryPoints.ts
Jonghakseo: This is an unnecessary assertion.

pages/content-ui/script/entryPoints.ts
```diff
@@ -0,0 +1,52 @@
...
+      .map(item => item.name);
+
+  foldersToProcess.forEach(folderName => {
+    const shouldExclude = options.exclude && options.exclude.includes(folderName);
+
+    if (!shouldExclude) {
+      const entryFolder = resolve(srcDir, folderName);
+      const entryFileTS = resolve(entryFolder, 'index.ts');
+      const entryFileTSX = resolve(entryFolder, `${folderName}.tsx`);
+      const existingEntryFileTSX = resolve(srcDir, 'App.tsx');
```
Jonghakseo, line 23: The details of creating an existing App.tsx as an entry point for pages is a very specific implementation that is difficult to understand from the outside.

App.tsx is not currently used anywhere, so it seems like a good idea to rename it to something like Template.tsx.

chrome-extension/manifest.js
```diff
@@ -51,7 +51,7 @@ const manifest = deepmerge(
       },
       {
         matches: ['http://*/*', 'https://*/*', '<all_urls>'],
-        js: ['content-ui/index.iife.js'],
+        js: ['content-ui/page1/index_dev.js'],
```
Jonghakseo, line 54: Because of missing implementation about shadow dom injection, this is totally not working right now

Jonghakseo (commented):
There are three changes to this PR

1. centralization of tailwindcss bundling
2. add multiple entry points for content-ui
3. update package (?)

First of all, the centralization of the CSS bundling process into the build process seems like a good idea. Basically I support your idea, but I'm concerned that we need to make sure there are no side effects or unexpected issues, that it's not a common implementation and users may struggle when adding the plugin from existing settings(by some tutorials, guide...), and that it's difficult to set the settings separately on each page.

Second, I'm skeptical that adding an entry point for content-ui is a necessary task. Users can achieve the same thing by copy/pasting the pages they want, such as content, content-ui, and registering them in the manifest if they want. My understanding is that this is a convenience feature, is that correct? Please let me know if I'm missing something.

.gitignore
```diff
@@ -23,4 +23,5 @@ chrome-extension/public/manifest.json
 **/tailwind-output.css
 
 # vite timestamp (because of bug from lib, remove it after fix will have been realased)
-**/vite.config.mts.timestamp-*
\ No newline at end of file
+**/vite.config.mts.timestamp-*
+pnpm-lock.yaml
```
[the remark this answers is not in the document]
tngflx: oops i forgot ive done that, will remove it.. i just have to push another commit right?

packages/hmr/lib/plugins/tailwindcss-bundler.ts
```diff
@@ -0,0 +1,42 @@
+import { Plugin } from 'vite';
+import fs from 'fs/promises';
+import path from 'path';
```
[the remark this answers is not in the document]
tngflx: ohh node must be included? what happen if don't

pages/content-ui/script/entryPoints.ts
tngflx: > This is an unnecessary assertion.

erm we only want folder name, and not file names, readdirsync will read filenames too which we don't want.

pages/content-ui/vite.config.mts
tngflx: hmm my repo worked fine, but im really puzzled how the test all failed, i didn't even touched the content-runtime stuff

pages/content-ui/vite.config.mts
tngflx: filename should be different in dev and production mode, perhaps hashes should be applied in production? that's the convention no?

pages/content-ui/src/page1/page1.tsx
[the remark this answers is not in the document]
tngflx: yeap i agree. i just want to show example

pages/content-ui/vite.config.mts
```diff
@@ -1,24 +1,38 @@
 import { resolve } from 'node:path';
-import { makeEntryPointPlugin } from '@extension/hmr';
 import { isDev, withPageConfig } from '@extension/vite-config';
+import { processCssPlugin } from '@extension/hmr';
```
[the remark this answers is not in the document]
tngflx: erm bro this is your own method ... haha you've written like this for your own plugin, so i just follow your convention :

import { makeEntryPointPlugin } from '@extension/hmr';
import { withPageConfig, isDev } from '@extension/vite-config';

const rootDir = resolve(__dirname);
const libDir = resolve(rootDir, 'lib');

export default withPageConfig({
  resolve: {
    alias: {
      '@lib': libDir,
    },
  },
  publicDir: resolve(rootDir, 'public'),
  plugins: [isDev && makeEntryPointPlugin()],
  
  i follow your makeentrypointplugin

pages/content-ui/vite.config.mts
tngflx: cause getentrypoints/makeentrypoints is resolving to the array of folder names to be used in vite.config entries. that's why its here. if not vite config don't know which folder to process. And yes this is more of convenience feature and people can add to manifest.json later

pages/content-ui/script/entryPoints.ts
tngflx: ahhh i just added that, cause user can choose between just single page or multi page, if user chooses to have multipage, he can always just add stuff into getentrypoints.includes which will generate those template for them. so its easier for them to understand.

pages/new-tab/src/index.tsx
tngflx: huh that's a great idea, perhaps will solve my build time error not found this file during build

chrome-extension/manifest.js
tngflx: erm bro.. basically what you need to do is just this :
```ts
import { createRoot } from 'react-dom/client';
import '@src/index.css';
import '../dist/global-out.css';
import NewTab from '@src/NewTab';
```

tngflx:
> There are three changes to this PR
> 
> 1. centralization of tailwindcss bundling
> 2. add multiple entry points for content-ui
> 3. update package (?)
> 
> First of all, the centralization of the CSS bundling process into the build process seems like a good idea. Basically I support your idea, but I'm concerned that we need to make sure there are no side effects or unexpected issues, that it's not a common implementation and users may struggle when adding the plugin from existing settings(by some tutorials, guide...), and that it's difficult to set the settings separately on each page.
> 
> Second, I'm skeptical that adding an entry point for content-ui is a necessary task. Users can achieve the same thing by copy/pasting the pages they want, such as content, content-ui, and registering them in the manifest if they want. My understanding is that this is a convenience feature, is that correct? Please let me know if I'm missing something.

based on ur first question, since the original code was already opinionated on tailwindcss i just help u to internalize the bundle process which is the exact same as running the build:tailwindcss script, your approach doesnt provide hmr and mine does. we can ofc work on different ui package which people ccan just install whtever ui library they want. but the step of bundling should be the same

.gitignore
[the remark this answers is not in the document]
PatrykKuniczak: Yeah, that's your normal workflow, PR it's only for tracking it

packages/hmr/lib/plugins/tailwindcss-bundler.ts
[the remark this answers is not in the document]
PatrykKuniczak: https://dev.to/yenyih/new-nodejs-protocol-imports-3pnf

pages/content-ui/vite.config.mts
PatrykKuniczak: Try to e.g. inject runtime script several times, you will have an error

pages/content-ui/vite.config.mts
PatrykKuniczak: No, i don't think it's really necessary to add prefix
If there's some differences between dev and prod versions, let's add if statement or sth inside code, and you'll have other output file.

PatrykKuniczak:
@tngflx @Jonghakseo Are you want to work on it?

I think it's very good idea to implement it, because of 
`multiple entryfiles just for content-ui` but i think it should accept multiple files for all kind of contents :)

PatrykKuniczak:
@Jonghakseo Let's take a look on first solution, about tailwind if it's necessary let's implement it in some way, because if i'm right issue still exist, but you're more familiar with it, i'm not following it much.

I can open PR with solution for handling multiple content scripts in the near future.

PR #773, opened 2024-10-13
Fix set storage before init cache

jackluson:

## Priority*

## Changes*
Add the inited state and make sure that the callback in the set function can get the correct cache value.


Jonghakseo (approved):
LGTM

PR #793, opened 2024-12-04
docs: fix access to env var using VITE_ prefix in README.md

dandv:
## Priority*

## Purpose of the PR*
In my previous large PR I forgot to add the `VITE_` prefix to accessing the env var via import.meta.env.

## Reference

https://vite.dev/guide/env-and-mode#env-files


dandv:
I haven't tested this *super* thoroughly, but it seems that *adding* an env var to `.env` requires restarting `pnpm dev` for the variable to be accessible via `import.meta.env.VITE...`. If that is indeed the case, it should be mentioned in that section.

Jonghakseo (approved):
LGTM

PR #819, opened 2024-12-22
chore: enable emptyOutDir for all Vite config files, optimize extension size

dorlanpabon:
# chore: enable emptyOutDir for all Vite config files, optimize extension size

**Description**  
This PR enables the `emptyOutDir` option in all Vite configuration files and optimizes the build settings to reduce the extension size. With this change, each build process will clean the output directory before generating new files, helping to avoid potential conflicts or stale artifacts and reducing the final extension size.

---

## Priority

---

## Purpose of the PR

- Enable the `emptyOutDir: true` option in all Vite-based build environments.  
- Ensure a clean output folder for generated artifacts to prevent stale files.  
- Optimize extension size.

---

## Changes

1. **pages/content-runtime/vite.config.mts**  
   - Added `emptyOutDir: true`.

2. **pages/content-ui/vite.config.mts**  
   - Added `emptyOutDir: true`.

3. **pages/content/vite.config.mts**  
   - Added `emptyOutDir: true`.

4. **pages/devtools-panel/vite.config.mts**  
   - Added `emptyOutDir: true`.

5. **pages/devtools/vite.config.mts**  
   - Added `emptyOutDir: true`.

6. **pages/new-tab/vite.config.mts**  
   - Added `emptyOutDir: true`.

7. **pages/options/vite.config.mts**  
   - Added `emptyOutDir: true`.

8. **pages/popup/vite.config.mts**  
   - Added `emptyOutDir: true`.

9. **pages/side-panel/vite.config.mts**  
    - Added `emptyOutDir: true`.

In total, 9 configuration files were updated to ensure that each build process generates a clean output directory and helps reduce the final extension size.

---

## How to Test the Feature

1. Run the build for each project (e.g., `pnpm run build` or `npm run build`, depending on your setup).  
2. Check the `dist` folder (or your configured output directory) to confirm that previous contents are removed before generating new files.  
3. Verify whether the extension size is reduced or remains as expected.  
4. Test each generated artifact to ensure that the extension and its various pages function correctly without leftover files.

---

## Reference

- [Vite Build Options](https://vitejs.dev/config/build-options.html#build-emptyoutdir)

If you have any questions or need more information, feel free to comment here. Thanks!

### DIFF
--- a/chrome-extension/vite.config.mts
+++ b/chrome-extension/vite.config.mts
@@ -34,7 +34,7 @@ export default defineConfig({
       fileName: 'background',
     },
     outDir,
-    emptyOutDir: true,
+    emptyOutDir: false,
     sourcemap: isDev,
     minify: isProduction,
     reportCompressedSize: isProduction,
--- a/chrome-extension/vite.config.mts
+++ b/chrome-extension/vite.config.mts
@@ -34,7 +34,7 @@ export default defineConfig({
       fileName: 'background',
     },
     outDir,
-    emptyOutDir: false,
+    emptyOutDir: true,
     sourcemap: isDev,
     minify: isProduction,
     reportCompressedSize: isProduction,
--- a/pages/content-runtime/vite.config.mts
+++ b/pages/content-runtime/vite.config.mts
@@ -19,5 +19,6 @@ export default withPageConfig({
       fileName: 'index',
     },
     outDir: resolve(rootDir, '..', '..', 'dist', 'content-runtime'),
+    emptyOutDir: true,
   },
 });
--- a/pages/content-ui/vite.config.mts
+++ b/pages/content-ui/vite.config.mts
@@ -21,5 +21,6 @@ export default withPageConfig({
       fileName: 'index',
     },
     outDir: resolve(rootDir, '..', '..', 'dist', 'content-ui'),
+    emptyOutDir: true,
   },
 });
--- a/pages/content/vite.config.mts
+++ b/pages/content/vite.config.mts
@@ -21,5 +21,6 @@ export default withPageConfig({
       fileName: 'index',
     },
     outDir: resolve(rootDir, '..', '..', 'dist', 'content'),
+    emptyOutDir: true,
   },
 });
--- a/pages/devtools-panel/vite.config.mts
+++ b/pages/devtools-panel/vite.config.mts
@@ -13,5 +13,6 @@ export default withPageConfig({
   publicDir: resolve(rootDir, 'public'),
   build: {
     outDir: resolve(rootDir, '..', '..', 'dist', 'devtools-panel'),
+    emptyOutDir: true,
   },
 });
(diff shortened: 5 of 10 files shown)


dorlanpabon:
## With `emptyOutDir: true`
image

When `emptyOutDir: true` is enabled, the final extension size remains stable because old build artifacts are removed before generating new files. This ensures you’re not accumulating previous compilations in the output folder and prevents the package size from ballooning over time.

---

## Without `emptyOutDir`
image

When `emptyOutDir` is disabled (or set to false), old build files remain in the output folder. Each new compilation adds more artifacts—often around 1MB per change—which can quickly inflate the package size (up to 300MB in some cases), significantly affecting the browser’s load performance for the extension.

Jonghakseo:
The danger is that the way the `emptyOutDir` option works is that it empties outDir before creating a new build file.

If a new build takes about a second, then existing scripts will get errors when referencing existing resources on refresh, etc.

The concern does not arise in production builds due to the `emptyOutDir: isProduction` option in `withPageConfig` and `pnpm clean:bundle` in the `pnpm build` command.

The problem in development environments can be solved by making sure that the first run of `pnpm dev` precedes an action like `pnpm clean:bundle`.

😄

dorlanpabon:
Thank you for your explanation, @Jonghakseo I understand your point about the potential issues in development environments caused by clearing the outDir before compiling, especially regarding existing resources during active development.

However, in my personal experience, I’ve encountered accumulated extension sizes of up to 300MB, and this isn’t from bundling but simply from running pnpm dev and making changes. Each change leaves behind artifacts, which over time negatively impacts the development experience, making the extension load progressively slower to the point where it becomes unsustainable for my workflow.

In my case, I implemented emptyOutDir: true and have worked with much cleaner and more manageable results. It’s true that I’ve experienced a minor inconvenience with the popup, where I have to close and reopen it after each change, but aside from that, everything works properly, and the overall experience is much smoother.

I truly appreciate your explanation and understand the nuances you’ve pointed out. Perhaps we could explore a hybrid solution that maintains the benefits of a clean environment without overly impacting the development experience.

Thank you again for your time and dedication!

Jonghakseo:
> Thank you for your explanation, @Jonghakseo I understand your point about the potential issues in development environments caused by clearing the outDir before compiling, especially regarding existing resources during active development.
> 
> However, in my personal experience, I’ve encountered accumulated extension sizes of up to 300MB, and this isn’t from bundling but simply from running pnpm dev and making changes. Each change leaves behind artifacts, which over time negatively impacts the development experience, making the extension load progressively slower to the point where it becomes unsustainable for my workflow.
> 
> In my case, I implemented emptyOutDir: true and have worked with much cleaner and more manageable results. It’s true that I’ve experienced a minor inconvenience with the popup, where I have to close and reopen it after each change, but aside from that, everything works properly, and the overall experience is much smoother.
> 
> I truly appreciate your explanation and understand the nuances you’ve pointed out. Perhaps we could explore a hybrid solution that maintains the benefits of a clean environment without overly impacting the development experience.
> 
> Thank you again for your time and dedication!

@dorlanpabon 

I completely understand what you are trying to say. As you say, there is indeed a possibility that you are experiencing a glitch and a solution seems to be needed. 

Specifically, maybe there is a way to clear the files used in the previous build right after the build instead of right before the build.

I really appreciate you sharing your personal experience and will try to implement your solution, thanks again.

PR #834, opened 2025-01-21
Better-er support for firefox

### LINKED ISSUE #833: Duplicate add-on ID found when submitting to AMO
**Describe the bug**
When attempting to submit the addon to [AMO](https://addons.mozilla.org/en-US/developers/addon/submit), I get "**Duplicate add-on ID found**". I've tried updating the `name` (both by updating `messages.json`, and just hardcoding a name to a name that very likely isn't taken), and adding an `id` to the `manifest.js` (which ends up in the `manifest.json` inside the `xpi` file when running `pnpm zip:firefox`.

**To Reproduce**
1. Change name for the extension, either by updating `messages.json` or by hardcoding a name in `manifest.js`
2. Build the extension for firefox via `pnpm build:firefox`
3. Pack the extension for firefox via `pnpm zip:firefox`
4. Upload the `dist-zip/XXX.xpi` file to [AMO](https://addons.mozilla.org/en-US/developers/addon/submit)

**Expected behavior**
Extension should be able to be uploaded and signed.

**Screenshots**

Image

**Desktop (please complete the following information):**
 - OS: Windows
 - OS Version: Version	10.0.22631 Build 22631
 - Browser: Firefox
 - Browser Version: 134.0.1 (64-bit)
 - Node Version: v22.12.0
 - Other Necessary Packages Version:

nivekmai:
Firefox addons need to have a unique add-on ID. The add-on ID is specified via `browser_specific_settings.gecko.id` in the manifest. #616 consistently set the add-on ID to "example@example.com", which won't work for anyone else who wants to publish an add-on.

Make the add-on ID configurable via the core manifest.js, since chrome extensions don't seem to care about the extra manifest key.

## Priority

## Purpose of the PR
Makes it so the repository creates Firefox extensions that can actually be used (you must sign them to use them, and you must have a unique ID to sign them).

## Changes
Add the add-on ID to `manifest.js`

## How to check the feature
1. Change id for the extension in manifest.js
2. Build the extension for firefox via pnpm build:firefox
3. Pack the extension for firefox via pnpm zip:firefox
4. Upload the dist-zip/XXX.xpi file to [AMO](https://addons.mozilla.org/en-US/developers/addon/submit)

## Reference
Fixes #833


PatrykKuniczak:
But it has no sense.
You have deleted it from `impl.ts` which add it for Firefox and add to `manifest`, but create info for users to remove it when they want to use only chrome.

It seems to have no difference between previous version

nivekmai:
Generally, I didn't think people would go looking through impl.ts to configure their extension (I didn't). I would think that impl.ts can be for autogenerated configs.

I'd also think you could have just the id in manifest.js, and let impl.ts do the object wrapping/min version. 

But I disagree that you should expect users of the boilerplate to go dig through the manifest parser code to find where to set the ID (since it's required for manifest v3 add-ons).

PatrykKuniczak:
@nivekmai In that case, if you want to move it to `manifest.js` then let's move all values firefox specific, but create it as additional object, which need to be merge with `deepmerge` with the rest of the manifest config in with `__FIREFOX__  === 'true'` condition.

Then i can agree with this PR

Jonghakseo:
Hello @nivekmai. Thank you for your contribution. We're working on a rather large and important change in dev, and I've created a new PR to reflect this and your suggestion together. If you're happy with this, I'd like to keep your changes in dev and launch them in our 0.4.0.

- #845

Jonghakseo (approved):

PR #835, opened 2025-01-23
Fix: theme toggle each time content script is loaded

tr4cks:
> `*` denotes required fields

## Priority*

## Purpose of the PR*

Make sure the theme doesn't change every time the content script is loaded into a page. This can be a bit confusing the first time you test the project.

## Changes*

I removed the `toggleTheme()` function in the content script and the dependency in `package.json`. I added an external function to still show an example of how to define functions in an external module.

## How to check the feature

Open the extension's side panel and then search in the browser's search bar. The theme no longer changes every time you change pages (when the content script is loaded).

### DIFF
--- a/pages/content/package.json
+++ b/pages/content/package.json
@@ -19,8 +19,7 @@
     "type-check": "tsc --noEmit"
   },
   "dependencies": {
-    "@extension/shared": "workspace:*",
-    "@extension/storage": "workspace:*"
+    "@extension/shared": "workspace:*"
   },
   "devDependencies": {
     "@extension/hmr": "workspace:*",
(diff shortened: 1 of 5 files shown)


tr4cks:
is it normal that the update also affected another line?

**From:**
> expect-webdriverio: 5.0.2(@wdio/globals@9.4.5(@wdio/logger@`9.1.0`))(@wdio/logger@9.4.4)(webdriverio@9.4.5)

**To:**
> expect-webdriverio: 5.0.2(@wdio/globals@9.4.5(@wdio/logger@`9.4.4`))(@wdio/logger@9.4.4)(webdriverio@9.4.5)

*Other lines have been modified, but they also concern the same thing.*

**Let me know if I leave it the way it was or update that too.**

Thanks

PatrykKuniczak:
@tr4cks Probably it have `^` in version inside this package, and on package change `pnpm` pull latests version with minor changes , that's why it change.

I'm not very sure about that, but that's probably it, but after all, run `pnpm i` and push --force-with-lease here :)

PatrykKuniczak (approved):
LGTM

PR #837, opened 2025-01-25
fix: Vite Rollup error about resolving import with alias

PatrykKuniczak:
> `*` denotes required fields

## Priority*

## Purpose of the PR*
Fix ENOENT error about resolve import "@/lib/utils" 

## Changes*
I've remove alias from `tsconfig` and change import to relative in `Button.tsx`

## How to check the feature
Run `pnpm dev` or `pnpm build`

## Reference
Zrzut ekranu 2025-01-25 195116

### DIFF
--- a/packages/ui/lib/components/Button.tsx
+++ b/packages/ui/lib/components/Button.tsx
@@ -1,5 +1,5 @@
 import type { ComponentPropsWithoutRef } from 'react';
-import { cn } from '@/lib/utils';
+import { cn } from '../utils';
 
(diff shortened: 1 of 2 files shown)


Jonghakseo:
We need aliasing when using shadcn ui components. Is there any solution for this problem?

PatrykKuniczak:
@Jonghakseo Not for now, maybe you have any, but if not, we need to merge it, because we can't realase it with bug

Jonghakseo (approved):

Issue #477, opened 2024-05-21
Constantly getting error in console `WebSocket connection to 'ws://localhost:8081/' failed:`


IzaGz:
Open console and see this error:
`Reload server disconnected.
Please check if the WebSocket server is running properly on ws://localhost:8081. This feature detects changes in the code and helps the browser to reload the extension or refresh the current tab.`

<img width="520" alt="Снимок экрана 2024-05-22 в 01 43 07" src="https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/assets/19924220/3f257883-c282-4eba-9803-f2b3d3a9ef2e">

also in messages (not errors):
<img width="394" alt="Снимок экрана 2024-05-22 в 01 48 43" src="https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/assets/19924220/ee7e8ad2-fa0e-4071-84a4-0127f0321d5f">

This clogs the console, not convenient.
How to fix this?

Jonghakseo:
Plz check your dev-server

```bash
$ pnpm dev-server
```

This is automatic on `pnpm dev` in latest version

- https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/pull/467

IzaGz:
i am trying latest version , starting it as `pnpm dev` but still getting this error, 
i have tried to run the `pnpm dev` in separate terminal and getting error that this is already running

ulises-castro:
Same issue

IzaGz:
still getting this 
@PatrykKuniczak @Jonghakseo  Can you please, reaopen this issue?

<img width="1054" alt="Снимок экрана 2024-06-11 в 00 54 16" src="https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/assets/19924220/9ed1b2eb-5e50-419b-8114-e4d575674e68">

Jonghakseo:
Have you checked to see if there are any other processes already occupying port 8081, and would like some information about the operating environment (OS, node version, etc.)?

IzaGz:
@Jonghakseo there is no other processes running the 8081, only dev-server starting ont this port, macOS Sonoma 14.5 , 
node -v 20.13.1

Jonghakseo:
Have you tried reloading the extension?

IzaGz:
@Jonghakseo 
Yes! Of course

ulises-castro:
[...]

PR #852, opened 2025-01-30
fix/ws-reconnect-on-dev-shut-down

PatrykKuniczak:
> `*` denotes required fields

## Priority*

## Purpose of the PR*
I want to avoid displaying error message of ws from #477 issue

## Changes*
I've created func which edit IS_DEV var on dist/background.js file on node process termination, for avoid ws reconnecting after WSS server were stopped because of script interruption

## How to check the feature
1. Run `pnpm dev` or `pnpm dev:firefox`
2. Wait for the end of building process
3. Interrupt with e.g `CTRL + C` in terminal


Jonghakseo (changes requested):
1. `pnpm dev` doesn't run because `SIGKILL` signal is not available on POSIX-compliant (Unix, Linux, macOS)
2. Websocket server doesn't shut down with `pnpm dev` process
3. This solution doesn't seem to work for other content scripts, new-tab, etc. except background.

PatrykKuniczak:
@Jonghakseo Now it should work everywhere.

I've tested it and it's shutting down with `pnpm dev` all processes on 8081 are down after i've closed `dev`

Jonghakseo:
스크린샷 2025-02-16 오후 1 51 56

Jonghakseo:
This appears to be a timing issue. I was wondering if dist/background might not be present when the hmr module is running.

PatrykKuniczak:
@Jonghakseo I can't reproduce it.
Have you tried `pnpm clean:install` before?

If yes, then give me steps to reproduce.

Jonghakseo (commented):
스크린샷 2025-03-16 오후 8 09 15

There’s still a critical issue occurring on the second run of `pnpm dev` as described above.

I think the current implementation that reads the terminal signals is too risky. How about close this PR?

To be honest, I'm not sure whether the WebSocket connection error that occurs when HMR shuts down is a major issue during development. The same happens in other tools that support HMR, like Next.js and Vite.

PatrykKuniczak:
@Jonghakseo Yeah, i'm also have been thinking to close it.

That's really issue with WebSocket and they have done it with reason, i hope :)

PR #858, opened 2025-02-04
Fix: Ensure the output directory exists when the public directory is deliberately omitted

tr4cks:
> `*` denotes required fields

## Priority*

## Purpose of the PR*

Ensure the ability to delete `public` directories in projects located within the `pages` directory.
To explain the problem in detail: when the `public` directory is created, `Vite` prioritizes managing this directory and therefore creates the folder structure needed for the build before calling the `generateBundle()` function within the `makeEntryPointPlugin()` plugin. The issue is that when the `public` directory doesn't exist, the folder structure isn't created, and this plugin requires an existing directory structure to write to files.

## Changes*
Creating the required folder structure within the `makeEntryPointPlugin()` plugin if it doesn't exist.

## How to check the feature

Delete the `public` directory within a project, such as `content-ui`, run `npm run dev`, and check that there are no errors in the console for the tested project.

### DIFF
--- a/packages/hmr/lib/plugins/make-entry-point-plugin.ts
+++ b/packages/hmr/lib/plugins/make-entry-point-plugin.ts
@@ -18,6 +18,11 @@ export function makeEntryPointPlugin(): PluginOption {
         throw new Error('Output directory not found');
       }
 
+      // Ensure the output directory exists.
+      if (!fs.existsSync(outputDir)) {
+        fs.mkdirSync(outputDir, { recursive: true });
+      }
+
       for (const module of Object.values(bundle)) {
         const fileName = path.basename(module.fileName);
         const newFileName = fileName.replace('.js', '_dev.js');


tr4cks:
@Jonghakseo I’m not sure if there’s anything to do in this function or not:
`packages/hmr/lib/plugins/make-entry-point-plugin.ts`
```ts
    closeBundle() {
      cleanupTargets.forEach(target => {
        fs.unlinkSync(target);
      });
    },
```

PatrykKuniczak:
> @Jonghakseo I’m not sure if there’s anything to do in this function or not: `packages/hmr/lib/plugins/make-entry-point-plugin.ts`
> 
> ```ts
>     closeBundle() {
>       cleanupTargets.forEach(target => {
>         fs.unlinkSync(target);
>       });
>     },
> ```

It's necessary, it only looks like unused

tr4cks:
In fact, I wanted to know whether I should delete the directory I created during initialization. But I don't think it's a good idea since it can still be used by `Vite` later in the process. So I don't know. Or check if there are still files in it. I don't know how useful that would be.

PatrykKuniczak:
@tr4cks This plugin won't be exist anymore after #846

PR #867, opened 2025-02-10
Ensure valid manifest for Firefox when background or options are not present

PrinceGupta1999:
> `*` denotes required fields

## Priority*

## Purpose of the PR*
Improve handling for making compatible manifest for Firefox when Background Scripts or Options UI are not used in manifest.js 

## Changes*
Add if checks to ensure correct manifest.json for Firefox when `manifest.background?.service_worker` or `manifest.options_page` are undefined

## How to check the feature
1. Comment or remove `options_page` or `background` in `manifest.js`.
2. Run `pnpm build:firefox`
3. The resultant `dist/manifest.json` will be valid now, which was not the case before

### DIFF
--- a/packages/dev-utils/lib/manifest-parser/impl.ts
+++ b/packages/dev-utils/lib/manifest-parser/impl.ts
@@ -14,14 +14,18 @@ function convertToFirefoxCompatibleManifest(manifest: Manifest) {
     ...manifest,
   } as { [key: string]: unknown };
 
-  manifestCopy.background = {
-    scripts: [manifest.background?.service_worker],
-    type: 'module',
-  };
-  manifestCopy.options_ui = {
-    page: manifest.options_page,
-    browser_style: false,
-  };
+  if (manifest.background?.service_worker) {
+    manifestCopy.background = {
+      scripts: [manifest.background?.service_worker],
+      type: 'module',
+    };
+  }
+  if (manifest.options_page) {
+    manifestCopy.options_ui = {
+      page: manifest.options_page,
+      browser_style: false,
+    };
+  }
   manifestCopy.content_security_policy = {
     extension_pages: "script-src 'self'; object-src 'self'",
   };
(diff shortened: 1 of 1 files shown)


packages/dev-utils/lib/manifest-parser/impl.ts
```diff
@@ -14,14 +14,18 @@ function convertToFirefoxCompatibleManifest(manifest: Manifest) {
...
-    scripts: [manifest.background?.service_worker],
-    type: 'module',
-  };
-  manifestCopy.options_ui = {
-    page: manifest.options_page,
-    browser_style: false,
-  };
+  if (manifest.background?.service_worker) {
+    manifestCopy.background = {
+      scripts: [manifest.background?.service_worker],
```
PatrykKuniczak, line 19:
```suggestion
      scripts: [manifest.background.service_worker],
```

It doesn't need `?` because you already know this exists

PatrykKuniczak (changes requested):

PrinceGupta1999:
thanks @PatrykKuniczak applied the suggestion

PatrykKuniczak (approved):
LGTM

Thanks for your contribution, we'll merge it soon :)

Jonghakseo (approved):
LGTM

tr4cks:
**A problem with this PR:**

I agree that on Firefox the option page is mandatory, I'm not against adding it. The problem is that on Chrome it's not mandatory, so if you really want to do a project on Chrome and you don't declare the option page, this PR will still add an option page reference in the manifest that doesn't really exist and the extension won't load on Chrome. This will probably also be the case with Firefox when an option page does not exist (not tested).

Perhaps we need to find a more appropriate solution.

A solution could be to add a default html page only on firefox and only if an option page doesn't already exist (isn't already defined).

But why I'm not in favor of this kind of default value idea:
The developer is no longer really aware that a page has been created by default, and doesn't have full control over what's going on.

Personally, I'd let the developer handle this kind of detail, or even indicate with a warning that there may be a problem when the developer builds for Firefox and some files are missing.

Let me know what you think, in any case we need to be able to set no option page for chrome.

_As for the background script, I don't know how it works, I haven't tested it, but it could be the same problems._

PR #877, opened 2025-02-16
Update deps

Jonghakseo:
> `*` denotes required fields

## Priority*

## Purpose of the PR*

Updated many dependencies.

## Changes*

- Update deps
- Fixed some type errors


PatrykKuniczak (commented):
First of all, you shouldn't bump packages with other changes

Secondly, UI module was `module` with reason.

Let's try to import e.g `env` in `ui`, you'll see
image

Jonghakseo:
> First of all, you shouldn't bump packages with other changes
> 
> Secondly, UI module was `module` with reason.
> 
> Let's try to import e.g `env` in `ui`, you'll see 


I don't think env is in the UI package's dependencies, and I don't think it's a layer where env is typically used.

The development experience and UI of the react component on the page should be the same, and the .js extension when adding shadcn, etc. can cause an unpleasant integration experience.

PatrykKuniczak:
> > First of all, you shouldn't bump packages with other changes
> > 
> > Secondly, UI module was `module` with reason.
> > 
> > Let's try to import e.g `env` in `ui`, you'll see 
> 
> 
> I don't think env is in the UI package's dependencies, and I don't think it's a layer where env is typically used.
> 
> The development experience and UI of the react component on the page should be the same, and the .js extension when adding shadcn, etc. can cause an unpleasant integration experience.

Probably no, but if somebody want to use it, for E.g display devtools under condition, they will have problem.

Overall, in this case, we need to have docs which told to users, to don't use env or other packages which aren't familiar with `base` for avoid questions in the future

PR #891, opened 2025-03-15
feat/configure-eslint-for-esnext

PatrykKuniczak:
> `*` denotes required fields

## Priority*

## Purpose of the PR*
Create new eslint rules, about which we was talking on team meeting

## Changes*
I've declared more rules for imports and one for enforce using `const` instead of `var`, `let` and `function` to follow es6+ syntax  

## How to check the feature
Run e.g `pnpm dev`


eslint.config.ts
```diff
@@ -52,9 +52,23 @@ export default ts.config(
...
+        },
+      ],
+      'prefer-const': 'error',
+      'no-var': 'error',
+      'import-x/named': 'error',
+      'import-x/namespace': 'error',
+      'import-x/default': 'error',
+      'import-x/export': 'error',
+      'import-x/no-named-as-default': 'error',
+      'func-style': ['error', 'expression', { allowArrowFunctions: true }],
```
Jonghakseo: 👍🏼

Jonghakseo (approved):
The consistently well-written commit list made it easy to understand. 

I have a personal question—if you don’t mind me asking, what was the reason for fixing the `func-style` rule to `expression` among the newly added rules? Aside from maintaining consistency, I’m curious about what else you took into consideration.

LGTM

eslint.config.ts
PatrykKuniczak: I don't have much to consider, because we have https://eslint.org/docs/latest/rules/func-style
only `declaration` and `expression` and i prefer second, because it's more current, much fresh code using func as const instead of `functions` and i hope it's good to follow latest approaches in coding at all.

If it's satisfied you, let's resolve conversation, else feel free to ask more :)

Jonghakseo: Sounds good. I was just curious, and I think it’s great for consistency. :)

PR #895, opened 2025-03-16
docs: remove info about git configuration refactor a little `Getting started` points and add new with IDE/Editor settings

PatrykKuniczak:
> `*` denotes required fields

## Priority*

## Purpose of the PR*
I've wanted to remove not current info and add new about recommended IDE configuration.

## Changes*
I've edit Getting started section of README and remove version from React for avoid case of forget to bump version also here and other packages, excluding manifest, don't have version and it should look like this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ the build speed and development experience by using Vite and Turborepo.
 
 ## Features
 
-- [React19](https://reactjs.org/)
+- [React](https://reactjs.org/)
 - [TypeScript](https://www.typescriptlang.org/)
 - [Tailwindcss](https://tailwindcss.com/)
 - [Vite](https://vitejs.dev/) with [Rollup](https://rollupjs.org/)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -72,22 +72,29 @@ the build speed and development experience by using Vite and Turborepo.
...
+    - <b>WebStorm</b>:
+      - Installed [Prettier plugin](https://plugins.jetbrains.com/plugin/10456-prettier)
+      - Enabled `File | Settings | Languages & Frameworks | JavaScript | Code Quality Tools | ESLint`\
+      -> `Automatic ESLint configuration` and `Run eslint --fix on save`
+      - Enabled `File | Settings | Languages & Frameworks | JavaScript | Prettier`\
+      -> `Automatic Prettier configuration` and `Run on save`
+      - Optional, but useful `File | Settings | Tools | Actions on Save`\
+      -> `Optimize imports` and `Reformat code`
+
+<b>Then, depending on the target browser:</b>
```
Jonghakseo, line 97: Hmm... I'm wondering if discussing IDE settings might be out of scope for this boilerplate. 

I'm not sure if it's appropriate to rely on the UI of a specific IDE like `VSCode` or `WebStorm`. 
I'm also worried that in the future, people might start adding instructions for other IDEs to the README.

PatrykKuniczak: Maybe not, but take a look, we have husky, which running eslint and prettier before commit and because of that i think it's good to have support with these actions from IDE, and it can help VSC to recognize TS version of project.

Maybe we could place it in other place/file.

Have you any idea?

Jonghakseo: Well... in any case, being considerate can be helpful for someone, so I think it's good. However, the plugin links and menu descriptions here are inevitably tied to the IDE’s UI. I’m starting to wonder if it’s really necessary to go into this much detail. 

If there’s official documentation for these settings, how about just leaving a link to that instead?

PatrykKuniczak: Ok, there's
I had been thinking about that and i've forgot about that.
I'll change it in minutes :)

PatrykKuniczak: @Jonghakseo Ok, done.
There's not official docs about setting TS version in VSC, only stackoverflow thread, that's why i've decided to leave it like that.

Rest have been simplified, i hope it looks better now :)

PR #897, opened 2025-03-16
fix/module-manager

PatrykKuniczak:
> `*` denotes required fields

## Priority*

## Purpose of the PR*
I've refactored code to reduce redundancy(create generic func instead of using a lot of hardcoded)

## Changes*
I've split files into a several more(smaller) and reduce code redundancy

## How to check the feature
Run `pnpm module-manager` and check if it works properly


Jonghakseo:
스크린샷 2025-03-30 오후 2 48 22

I just run `pnpm clean && pnpm i && pnpm module-manager`

packages/module-manager/lib/runModuleManager.ts
```diff
@@ -36,12 +39,22 @@ const runModuleManager = async () => {
...
+
+  execSync('pnpm -F chrome-extension lint:fix', {
+    stdio: 'inherit',
+    cwd: resolve('..', '..'),
+  });
+
+  await sleep(1000);
+  execSync('git add .', {
+    stdio: 'inherit',
+    cwd: resolve('..', '..'),
```
Jonghakseo, line 56: Running git commands feels like it goes beyond the scope of the boilerplate. If `git add .` isn't used, we wouldn't need that hacky sleep code mentioned above either.

Jonghakseo (commented):
I left a few comments about my questions.

packages/module-manager/lib/const.ts
```diff
@@ -0,0 +1,61 @@
...
+      matches: ['http://*/*', 'https://*/*', '<all_urls>'],
+      js: [`content/index.iife.js`],
+    },
+  },
+  'content-ui': {
+    content_scripts: {
+      matches: ['http://*/*', 'https://*/*', '<all_urls>'],
+      js: [`content-ui/index.iife.js`],
+    },
+  },
```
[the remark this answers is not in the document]
PatrykKuniczak: We'll add `.forEach()` inside https://github.com/Jonghakseo/chrome-extension-boilerplate-react-vite/blob/fix/module-manager/packages/module-manager/lib/utils.ts#L24

packages/module-manager/lib/runModuleManager.ts
```diff
@@ -36,12 +39,22 @@ const runModuleManager = async () => {
...
+  });
+
+  execSync('pnpm -F chrome-extension lint:fix', {
+    stdio: 'inherit',
+    cwd: resolve('..', '..'),
+  });
+
+  await sleep(1000);
```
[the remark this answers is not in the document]
PatrykKuniczak: I was trying a couple workarounds but only that works for me, if you're able you can try to fix it in other way, i'll be glad :)

packages/module-manager/lib/runModuleManager.ts
PatrykKuniczak: Why i've decided to move it here?
To ensure users won't forget to add it to tracking, we have of course new info in README, but i've thought that's good to do it automatically.

But i'll remove it

PatrykKuniczak:
> 스크린샷 2025-03-30 오후 2 48 22
> I just ran `pnpm clean && pnpm i && pnpm module-manager`

You have to build project with `pnpm build`, i've added info on README in:
d7bab3c436efb11a0ee6985f66dbd3d1feded9b0

README.md
```diff
@@ -138,6 +138,10 @@ can use only `content-script` without `@extension/` prefix
 
 ## How do I disable modules I'm not using?
 
+> [!IMPORTANT]
+> Make sure you have built project before
+> `pnpm build` or `pnpm build:firefox`
```
Jonghakseo, line 143: It appears that module-manager is likely the first command users will run right after cloning or forking the boilerplate. Requiring them to run pnpm build seems a bit unusual. Is there another solution?

PatrykKuniczak: I need to move `streamFileToZip` again inside package, i've moved it to `dev-utils` to avoid deduncancy.

But if we want to do it without build, i need to move it back inside `module-manager`

PatrykKuniczak: But yeah, it should work without additional build, if you don't have other idea, i need to move it back inside module.

PatrykKuniczak:
@Jonghakseo Everything fixed, ready for re-review

README.md
PatrykKuniczak: I've fixed it with additional `postinstall` script on `module-manager`

PR #932, opened 2025-04-04
fix/eslint-import-order

PatrykKuniczak:
> `*` denotes required fields

## Priority*

## Purpose of the PR*
I need to fix eslint conflict between rules for imports restriction

## Changes*
I've changed rules for `import-x/order`

## How to check the feature
Run `pnpm lint` or `pnpm lint:fix`


PatrykKuniczak:
@Jonghakseo Why merge doesn't wait for workflow(actions) end?

Previously it work with `auto-merge` and merge only if workflow was successful, are you change sth?

PR #952, opened 2025-04-21
feat(ui): tailwind native dark mode

niko-38500:
> `*` denotes required fields

## Priority*

## Purpose of the PR*
Allow to use the native tailwind dark mode modifier "dark:*" for a better DX

## Changes*
using chrome.storage to store the current theme and monitor the change to toggle de "dark" class on the <html> html tag, if not set the user system preference is used.

Added a react entry-point initializer to avoid repeat  

## How to check the feature
add a class with the dark modifier and check that the ui take the right tailwind class depending on the current theme

image


.gitignore
```diff
@@ -23,3 +23,7 @@ tsconfig.tsbuildinfo
 
 # compiled
 **/tailwind-output.css
+
+.pnpm-store
+
+eslint.config.js
```
PatrykKuniczak, line 29: it doesn't exits

packages/ui/lib/components/ToggleButton.tsx
```diff
@@ -1,21 +1,25 @@
...
 
 export const ToggleButton = ({ className, children, ...props }: ToggleButtonProps) => {
-  const { isLight } = useStorage(exampleThemeStorage);
+  const onClick = async () => {
+    chrome.storage.local.get(['theme']).then(({ theme }) => {
+      chrome.storage.local.set({
+        theme: theme === 'dark' ? 'light' : 'dark',
+      });
+    });
+  };
```
PatrykKuniczak, line 13: Why haven't you use exampleThemeStorage?

pages/devtools-panel/src/index.tsx
```diff
@@ -1,15 +1,5 @@
...
-  if (!appContainer) {
-    throw new Error('Can not find #app-container');
-  }
-  const root = createRoot(appContainer);
-
-  root.render(<Panel />);
-};
-
-init();
+initReactEntrypoint(<Panel />);
```
PatrykKuniczak, line 5: That's good, but place it in other PR and in `shared` and use it for each page

pages/devtools-panel/src/Panel.tsx
```diff
@@ -1,43 +1,54 @@
...
           Edit <code>pages/devtools-panel/src/Panel.tsx</code>
         </p>
-        <ToggleButton onClick={exampleThemeStorage.toggle}>{t('toggleTheme')}</ToggleButton>
+        <ToggleButton>{t('toggleTheme')}</ToggleButton>
       </header>
     </div>
   );
 };
 
 const ToggleButton = (props: ComponentPropsWithoutRef<'button'>) => {
```
PatrykKuniczak, line 35: Why have you redeclared ToggleButton?

PatrykKuniczak (changes requested):
Avoid duplication `ToggleButton`

And also `initReactEntrypoint` (but do it in other PR)

pages/devtools-panel/src/Panel.tsx
niko-38500: It was already redeclared. I'll replace with the existing one

PatrykKuniczak: Oh yeah, indeed.

Are you able to remove it in other PR.

I want to suggest you to enable dark mode, only it, here.
1 PR with helper for init react for each page
1 PR for remove redundancy for ToggleButton(use this one from `/ui/lib/components`) :)

niko-38500:
I've resolved your comments, I'll do other PR's as you have suggested for the removal of the duplicate toggle button and the creation of the react entrypoint I'll also make a PR to add a method in the example theme storage for the on changed because subscribe does not fit the use case

pages/side-panel/src/index.tsx
```diff
@@ -1,13 +1,30 @@
...
+
+    document.querySelector('html')?.classList.toggle('dark', theme === 'dark');
+  };
+
+  exampleThemeStorage.get().then(themeState => toggleTheme(themeState.theme));
+
+  chrome.storage.local.onChanged.addListener(changes => {
+    toggleTheme(changes['theme-storage-key'].newValue.theme as ThemeState['theme']);
+  });
+
```
PatrykKuniczak, line 21: It's really necessary?

I remember theme storage, set automatically default theme(basing on browser)

niko-38500: I don't think so, there is a fallback but we need to set the class of the html tag with this way to manage the dark mode

PatrykKuniczak: Maybe return from current implementation string with theme e.g `dark` and then in yeah page assign it.

niko-38500: I'm not sure to understand what you mean

niko-38500: I don't think it is possible since you can't use hooks outside of a component, but I can use the isLight property inside the storage instead of the theme property

PatrykKuniczak: Yeah, you can't use, but you can get document inside of component and inside of useStorage(and there you can assign it)

niko-38500: Ok, so I think I will do an HOC to handle this in my other pr to mutualize the react entrypoint. Is it ok for you ?

PatrykKuniczak: Ok, but that PR will need to be merged before that.

But feel free to do it in your way :)

niko-38500: What if we leave it like that for this pr and in the other PR I will create an HOC that handle the class switching on the html document ? It will be more clean in my opinion because handle this use case in a component induces a lot of code duplication. What is your thought ?

niko-38500: Nice :), I pushed the resolution of the conflict I rebased the base branch, I let you resolve the conversations and merge the PR or I can do it ? I'm not aware of your contributions policy

PatrykKuniczak: @niko-38500 You can resolve it, if you think it's resolved :)

If you have applied requested changes, feel free to resolve it.

pages/devtools-panel/src/Panel.tsx
PatrykKuniczak: @niko-38500 I think you can also remove `onClick` from all buttons props and place it inside component in second(from list above) PR

PatrykKuniczak: My bad, it's already inside XD
You need to only remove redundant one, and use it from `/ui/components`
But put translation inside ToggleButton and use it like:
`<ToggleButton/>`

pages/devtools-panel/src/index.tsx
```diff
@@ -1,8 +1,24 @@
...
+
+    document.querySelector('html')?.classList.toggle('dark', theme === 'dark');
+  };
+
+  exampleThemeStorage.get().then(themeState => toggleTheme(themeState.theme));
+
+  chrome.storage.local.onChanged.addListener(changes => {
+    toggleTheme(changes['theme-storage-key'].newValue.theme as ThemeState['theme']);
+  });
+
```
PatrykKuniczak, line 21: Remove it also.

pages/new-tab/src/index.tsx
```diff
@@ -1,8 +1,24 @@
...
+
+    document.querySelector('html')?.classList.toggle('dark', theme === 'dark');
+  };
+
+  exampleThemeStorage.get().then(themeState => toggleTheme(themeState.theme));
+
+  chrome.storage.local.onChanged.addListener(changes => {
+    toggleTheme(changes['theme-storage-key'].newValue.theme as ThemeState['theme']);
+  });
+
```
PatrykKuniczak, line 21: also

pages/options/src/index.tsx
```diff
@@ -1,13 +1,30 @@
...
+
+    document.querySelector('html')?.classList.toggle('dark', theme === 'dark');
+  };
+
+  exampleThemeStorage.get().then(themeState => toggleTheme(themeState.theme));
+
+  chrome.storage.local.onChanged.addListener(changes => {
+    toggleTheme(changes['theme-storage-key'].newValue.theme as ThemeState['theme']);
+  });
+
```
PatrykKuniczak, line 21: also

pages/popup/src/index.tsx
```diff
@@ -1,8 +1,24 @@
...
+
+    document.querySelector('html')?.classList.toggle('dark', theme === 'dark');
+  };
+
+  exampleThemeStorage.get().then(themeState => toggleTheme(themeState.theme));
+
+  chrome.storage.local.onChanged.addListener(changes => {
+    toggleTheme(changes['theme-storage-key'].newValue.theme as ThemeState['theme']);
+  });
+
```
PatrykKuniczak, line 21: same

pages/side-panel/src/index.tsx
PatrykKuniczak, line 21: same

PatrykKuniczak (changes requested):
Because toggle theme not working at all, i think it's good idea to implement working solution in this PR.

Let's edit toggleTheme to work and remove toggleTheme from all components.
And rest should be done in other PR

niko-38500:
> Because toggle theme not working at all, i think it's good idea to implement working solution in this PR.
> 
> Let's edit toggleTheme to work and remove toggleTheme from all components. And rest should be done in other PR

I found the issue with theme toggling, you made a pr that remove the tailwind base config from all the pages, but I manage the dark theme with a tailwind config in the base config file, why have you perform these changes ? I think it might be useful to have a global tailwind config

PatrykKuniczak:
> > Because toggle theme not working at all, i think it's good idea to implement working solution in this PR.
> > Let's edit toggleTheme to work and remove toggleTheme from all components. And rest should be done in other PR
> 
> I found the issue with theme toggling, you made a pr that remove the tailwind base config from all the pages, but I manage the dark theme with a tailwind config in the base config file, why have you perform these changes ? I think it might be useful to have a global tailwind config

I've removed it, because i want to not redefining it, it seems to be redundant.
If it's necessary, let's back those changes

PR #966, opened 2025-04-27
feat(docker): add docker support

niko-38500:
## Priority*

## Purpose of the PR*
add the support for building the extension with docker

update the README.md

## Changes*

add a docker compose file a Dockerfile and an entry

## How to check the feature

Read the readme.md changes to know how to test the PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@
 - [Getting started](#getting-started)
     - [Chrome](#getting-started-chrome)
     - [Firefox](#getting-started-firefox)
+- [With docker](#with-docker)
 - [Install dependency](#install-dependency)
     - [For root](#install-dependency-for-root)
     - [For module](#install-dependency-for-module)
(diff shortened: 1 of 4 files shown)


PatrykKuniczak:
@niko-38500 I have been thinking about that :)
I'll check it ASAP

PatrykKuniczak:
@niko-38500 I'm not sure, if this is really necessary.

I mean it's great idea, which i had a while ago, but for full-stack(e.g Nest + This teamplate) template.

But it seems to be overkill for simple template like us.
It seems to be overengeneering.

I'm confused.

I have plan to create full-stack template, and there i want to place docker, but i'm not sure if it's really necessary here.

Let's wait for @Jonghakseo vote.

PatrykKuniczak:
And there's big issue because of lack of the /dist in root

Also lack of rimraf, which need to be installed globally inside container:
image

IMO It's not good idea to complicate template more, it's complicated a lot now.

I think it's good idea to implement it in the future in other template, which i want to implement in the future, today i was writing about it in our [Discord](https://discord.gg/RhETcmdW)

PR #994, opened 2025-05-19
Module manager test

### LINKED ISSUE #937: Add tests for module-manager
**Is your feature request related to a problem? Please describe.**
Implement tests and add it for current workflow (GH Actions)

**Describe the solution you'd like**
Find a way to implement tests.
It could be run like: 
`pnpm module-manager -r popup && pnpm e2e:manager -- -r popup` (after #936) or sth like that.

govza:
> `*` denotes required fields

## Priority*

## Purpose of the PR*
Run tests for separated features

## Changes*
Created `e2e-moduler` GH workflow

## How to check the feature
1. pnpm module-manager --de tests random-module
2. pnpm run e2e
3. pnpm module-manager -r random-module
4. pnpm run e2e

Closes #937

### DIFF
--- a/.github/workflows/e2e-modular.yml
+++ b/.github/workflows/e2e-modular.yml
@@ -1,0 +1,71 @@
+name: Modular E2E Tests Matrix
+
+on: [ pull_request_target ]
+
+permissions:
+  contents: read
+  
+env:
+  ALL_SCENARIOS: '["content", "content-ui", "content-runtime", "background", "new-tab", "popup", "devtools", "side-panel", "options"]'
+
+jobs:
+  e2e-tests:
+    name: '${{ matrix.browser }} | Scenario: ${{ matrix.scenario }}'
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        browser: ['chrome', 'firefox']
+        scenario: ['content', 'content-ui', 'content-runtime', 'background', 'new-tab', 'popup', 'devtools', 'side-panel', 'options']
+
+    env:
+      CEB_CI: true
+      CURRENT_SCENARIO: ${{ matrix.scenario }}
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: pnpm/action-setup@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version-file: '.nvmrc'
+          cache: pnpm
+
+      - run: pnpm install --frozen-lockfile --prefer-offline
+
+      - name: Configure modules | scenario "${{ matrix.scenario }}"
+        run: |
+          echo "Deleting all features except '${{ matrix.scenario }}'"
+          pnpm module-manager --de ${{ matrix.scenario }} tests
+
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e-modular.yml
```diff
@@ -0,0 +1,70 @@
+name: Modular E2E Tests Matrix
+
+# on: [ pull_request_target ]
```
PatrykKuniczak, line 3: Remove it

.gitignore
```diff
@@ -23,3 +23,6 @@ tsconfig.tsbuildinfo
 
 # compiled
 **/tailwind-output.css
+
+# module-manager archives
+packages/module-manager/archive/**.*
```
PatrykKuniczak, line 28: It need to be on repo, because when somebody have removed 1 module, somebody else will need to be able to recover it.

PatrykKuniczak (changes requested):
What determines the current browser for `pnpm e2e` ?

Browser on which we're currently are or what?

And tests run will iterate on both browsers and run tests for both?

tests/e2e/specs/page-content-runtime.test.ts
PatrykKuniczak, line 1: I've forgot about it, good job :)

govza:
> What determines the current browser for `pnpm e2e` ?
> 
> Browser on which we're currently are or what?
> 
> And tests run will iterate on both browsers and run tests for both?
> 
> And fix CodeQL issues.
> 
> And it isn't possible to run it on local machine, only on CI, are you think about to add it also as a package.json script?

pnpm e2e runs by default on chrome, pnpm e2e:firefox is run on firefox
the matrix strategy defined in workflow allows to run them altogether in parallel,
It's possible to run it locally since it's just - remove all module, - run e2e, - restore random module, - run e2e,
but I don't see any use-case for it.

PatrykKuniczak:
@govza I'm also wondering about use case of local run, and i think we can skip it, because it should work only on CI.
  
  And why it isn't working here, i don't freaking know, i was trying everything to get knowledge, why it can't work.
  I see you have `pull_request_target` and i think it should work, let's set `permissions: write-all` for test, and check it then.
  
  But overall there's something wrong with settings of repo, because on my fork it also works better.
  @Jonghakseo need to take a look on that problem deeper.

.github/workflows/e2e-modular.yml
PatrykKuniczak, line 1: Are you checked scenario, where user deleted one or more modules and pushes `archive` folder to repo.
And this action was run without several modules, successfully?

govza: Kind of, basically the scenario is:
1           pnpm module-manager --de for-each-feature
2           pnpm run e2e
3           pnpm module-manager -r "$random_feature"
4           pnpm run e2e again

I think it's pretty good coverage not full of course, since for all possible combinations it will be 2^n -1, i.e 511 variants for 9 features, hence the random tests (which normally should be avoided).

PatrykKuniczak (approved):
LGTM very good work :)

